PR #4, opened 2019-05-30
Document thumbnail according to code

quite:


quite:
Aaah, scratch this... The usage in scrot.txt differs from the usage in options.c... Extraction should perhaps be automated somehow.

Issue #11, opened 2019-06-10
Several warnings when building the future version


eribertomota:
Hi all,

When preparing the 0.10 version, I can  see that source code produces several warnings when building over Debian Sid.

I attached a log.

@daltomi Can you take a look over it?

Cheers,

Eriberto

[build.zip](https://github.com/resurrecting-open-source-projects/scrot/files/3270290/build.zip)

daltomi:
I do not know what happens with git, but I can not create a PR without sending many changes.
Patch:
[fix_buid_warnings.patch.zip](https://github.com/resurrecting-open-source-projects/scrot/files/3270564/fix_buid_warnings.patch.zip)

daltomi:
The correct branch for this fix es #14

eribertomota:
Fixed by 7a3e323

PR #13, opened 2019-06-10
Fix  issue #11

daltomi:
Autotools say:  "Don’t make the mistake of including config.h in a public header file if  your project installs libraries and header files as part of your product set."  and that mistake was made by giblib.


daltomi:
Bad branch, close.

PR #14, opened 2019-06-10
Fix build warnings.

daltomi:
Autotools say: "Don’t make the mistake of including config.h in a public header file if your project installs libraries and header files as part of your product set." and that mistake was made by giblib.

Add prefix SCROT_ to config.h

Evaluate the return of system() call.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4,6 +4,7 @@ AC_INIT([scrot], [0.10], [https://github.com/resurrecting-open-source-projects/s
 AC_CONFIG_SRCDIR([src/main.c])
 AM_INIT_AUTOMAKE
 AC_CONFIG_HEADER([src/config.h])
+AX_PREFIX_CONFIG_H([src/scrot_config.h])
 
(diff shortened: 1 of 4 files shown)


daltomi:
Hi @eribertomota 

If you agree with this commit, I merge it, so I only just have to send one more commit to #15

Regards.

eribertomota:
Considering several changes and fixes, I will release the 1.0 version. Do you agree? If not, please, revert my last commit.

daltomi:
If I agree with the release.

The error does not happen in archlinux.
Here: https://github.com/facebook/proxygen/issues/12
"Looks like I missed autconf-archive package. Installing it solved the issue."

It is up to you because you use Debian.
In any case, issue #11 is not so serious, and the system() call problem can be corrected later if you decide to close this commit.

Regards.

eribertomota:
Worked fine with autoconf-archive. No more warnings. I will add a section about build dependencies in README file. I will merge.

PR #19, opened 2019-06-23
Added a check for autoconf-archive presence

eribertomota:
This check will avoid mistakes from users.


eribertomota:
Thanks!

Em domingo, 23 de junho de 2019, Daniel T. Borelli <notifications@github.com>
escreveu:

PR #20, opened 2019-07-05
adds option to embed hostname in output file name

jda:
I move between multiple computers and save screenshots to a folder that is synchronized between them with Syncthing. This is handy when I'm writing documentation on one computer and taking screenshots of whatever I'm documenting on the other. 
Problem: how to tell which screenshot came from what system?

This PR adds a `$a` format flag that adds the system's hostname to the file name.

I picked "a" because hostnAme (h and n were already in use).

### DIFF
--- a/man/scrot.1
+++ b/man/scrot.1
@@ -114,6 +114,7 @@ following specifiers are recognised:
 .PP
 .nf
 .fam C
+    $a  hostname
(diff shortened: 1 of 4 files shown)


eribertomota:
Thanks a lot for your implementation.

Issue #22, opened 2019-07-30
resurrect/retire giblib as well?


mathstuf:
Thanks for bringing this back. However, it depends on `giblib` which is still abandoned. Should this repo remove the dependency or should the org bring back `giblib` as well?

eribertomota:
Thanks for your message @mathstuf.

@daltomi What you think about it?

Regards,

Eriberto

jda:
I noticed that the giblib string and mem functions are [very similar to this file](https://chromium.googlesource.com/chromiumos/third_party/ktop/+/refs/heads/master/eprintf.c) in Chromium, which is attributed to a book:

> Excerpted from 'The Practice of Programming' by Brian W. Kernighan and Rob Pike

Anyway, I have a branch where I use it instead of the giblib string functions: https://github.com/jda/scrot/commit/6b81b38cf7b358575aa020173e424fa30173d36e
Will look at the gib_imlib functions next.

eribertomota:
Em seg, 5 de ago de 2019 às 02:29, Jade Auer <notifications@github.com>
escreveu:

> I noticed that the giblib string and mem functions are very similar to
> this file
> <https://chromium.googlesource.com/chromiumos/third_party/ktop/+/refs/heads/master/eprintf.c>
> in Chromium, which is attributed to a book:
>
> Excerpted from 'The Practice of Programming' by Brian W. Kernighan and Rob
> Pike
>
> Anyway, I have a branch where I use it instead of the giblib string
> functions: jda@6b81b38
> <https://github.com/jda/scrot/commit/6b81b38cf7b358575aa020173e424fa30173d36e>
> Will look at the gib_imlib functions next.
>
>
>
Thanks Jade!

mathstuf:
What's the progress here? Can an external-giblib-free release be made within the next few months?

daltomi:
Hi.
My first attempt, to **embed** it in `scrot` was wrong: #24 

It is true that it is discontinued, but replacing it in `scrot` I do not think necessary, it is much "reinvent the wheel" and as far as I know it has no bugs.
In any case I like the idea that may be in the `resurrecting-open-source-projects`

If someone does want to replace it, here is a list of the functions that are used from `scrot`:
```
[gib_list.c]
gib_list_*
	<build-in> 508 LOC

Note: scrot calls 2 functions:
	gib_list_add_end
	gib_list_length

[gib_utils.c]
gib_eprintf
	fflush
	fprintf
	va_start
[...]

PR #24, opened 2019-08-05
Include the giblib project source files.

daltomi:
Hi all,

There are approximately ~30 calls to giblib functions. So I think it is best to include it in the scrot project that is reinventing the wheel.

This PR add `src/giblib`  folder with the sources files and standard autotools files: README, AUTHORS, etc.

Regards.


eribertomota:
Hi @daltomi,

Sorry, but I fully disagree. Embedded libraries require extra effort to security maintenance or normal maintenance. It is strongly abominated in Debian. Our Lintian system always points it as an error. Please, see the following texts:

https://www.debian.org/doc/debian-policy/ch-source.html#convenience-copies-of-code
https://wiki.debian.org/UpstreamGuide#No_inclusion_of_third_party_code
https://lintian.debian.org/tags/embedded-library.html

Thanks!

Cheers,

Eriberto

daltomi:
You're right, thanks for indicating this.
Cheers.

eribertomota:
Thank you!

Em seg, 5 de ago de 2019 às 20:04, Daniel T. Borelli <
notifications@github.com> escreveu:

mathstuf:
@eribertomota In Fedora at least, giblib is *only* used by scrot. Upstream is dead and, AFAIK, nothing else uses it, so this might be a reasonable solution anyways.  Is giblib used in any other capacity in Debian?

jda:
For what it's worth, giblib has already been vendored by other projects. For example, feh: https://github.com/derf/feh/blob/master/src/gib_imlib.h

Looking at how scrot uses giblib I don't think we're getting that much from it, so we may as well remove it entirely.

eribertomota:
Hi folks,

@mathstuf, in Debian scrot and gozer depend of giblib.

@jda, can you try remove?

Cheers,

Eriberto

jda:
@eribertomota, yes :)
Only part that remains in my branch is the linked list handling. I'm trying to replace that with `sys/queue.h` (more discussion on #22 )

Issue #26, opened 2019-08-18
warnings with clang on NetBSD 9.99.4/amd64


ng-0:
Hi,

while evaluating this as an update for graphics/scrot in pkgsrc, I've got these warnings (on NetBSD 9.99.4 with clang + llvm 8). Fixes should be simple.

```
Making all in src
/usr/bin/make  all-am
clang -DHAVE_CONFIG_H -I.  -g -O3 -Wall -I/usr/X11R6/include   -I/usr/pkg/include -I/usr/pkg/include -I.  -DPREFIX=\""/usr/pkg"\" -I/usr/pkg/include -I/usr/pkg/include/giblib -I/usr/pkg/include  -I/usr/pkg/include -I/usr/include -I/usr/X11R7/include/freetype2 -I/usr/X11R7/include  -O2 -I/usr/pkg/include -I/usr/include -I/usr/X11R7/include/freetype2 -I/usr/X11R7/include -MT main.o -MD -MP -MF .deps/main.Tpo -c -o main.o main.c
mv -f .deps/main.Tpo .deps/main.Po
clang -DHAVE_CONFIG_H -I.  -g -O3 -Wall -I/usr/X11R6/include   -I/usr/pkg/include -I/usr/pkg/include -I.  -DPREFIX=\""/usr/pkg"\" -I/usr/pkg/include -I/usr/pkg/include/giblib -I/usr/pkg/include  -I/usr/pkg/include -I/usr/include -I/usr/X11R7/include/freetype2 -I/usr/X11R7/include  -O2 -I/usr/pkg/include -I/usr/include -I/usr/X11R7/include/freetype2 -I/usr/X11R7/include -MT getopt.o -MD -MP -MF .deps/getopt.Tpo -c -o getopt.o getopt.c
getopt.c:571:30: warning: implicitly declaring library function 'strcmp' with
      type 'int (const char *, const char *)' [-Wimplicit-function-declaration]
      if (optind != argc && !strcmp(argv[optind], "--"))
                             ^
getopt.c:571:30: note: include the header <string.h> or explicitly provide a
      declaration for 'strcmp'
getopt.c:649:15: warning: implicitly declaring library function 'strncmp' with
      type 'int (const char *, const char *, unsigned long)'
      [-Wimplicit-function-declaration]
         if (!strncmp(p->name, nextchar, nameend - nextchar))
              ^
getopt.c:649:15: note: include the header <string.h> or explicitly provide a
[...]

PR #27, opened 2019-08-18
Fix warnings in getopt.c

fraggerfox:
The patches applied to pkgsrc to fix the warnings in `getopt.c` is shared here.

1. Fix the missing `string.h` when building in NetBSD

    It seems `string.h` was conditionally included and was causing the compiler to spit out warnings. `string.h` is now unconditionally included to prevent this issues.

2. Fix the ambiguous `else` for the `if` in line 689

    This has been fixed by adding the correct braces to the `if` condition in line 689.

This pull request addresses #26 in the issue tracker.

### DIFF
--- a/src/getopt.c
+++ b/src/getopt.c
@@ -76,6 +76,8 @@
 #endif
 #endif
 
+#include <string.h>
+
 #ifndef _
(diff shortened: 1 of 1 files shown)


eribertomota:
Thanks!

Issue #30, opened 2019-08-19
scrot writes to a random-ass file name instead of the given one when it exists


phiresky:
I guess since #12 / #15, scrot ignores the given file name and instead silently writes the result to a completely different file.

To reproduce:

```
$ scrot /tmp/screenshot.png
$ scrot /tmp/screenshot.png
$ ls /tmp/*.png
screenshot_001.png????
```

scrot should either fail with an exit code if the file already exists, or override the destination file by default like any other sane non-wimpy unix tool. Appending some random number to the file name may be useful if no output file name is explicitly given, but why would I even give it an explicit filename if it just ignores it sometimes? If you really want to have the behaviour that currently is the default, there should in my opinion be a flag like `--no-clobber` as is the standard (see e.g. `man mv` or `man cp`).

This change breaks by [workflow](https://xkcd.com/1172/).

phiresky:
Ok, actually I use `gnome-screenshot -a -f` now (which also works without gnome), has this behaviour, and doesn't have the rectangle redrawing bug scrot has, so if no one else feels this way feel free to close this issue :)

aureooms:
I am observing a weirder behavior on my machine.

```
$ pacman -Ss scrot
community/scrot 1.2-1 [installed]
    Simple command-line screenshot utility for X
```

The following creates two files
```
$ ls /tmp/*.png
"/tmp/*.png": No such file or directory (os error 2)
$ scrot "$(mktemp --tmpdir "${USER}-system.lock-screen.XXXXXXXX.png")"
$ ls -la /tmp/*.png
.rw-------    0 aureooms 14 Sep 11:29 /tmp/aureooms-system.lock-screen.Daefkf83.png
.rw-r--r-- 185k aureooms 14 Sep 11:29 /tmp/aureooms-system.lock-screen.Daefkf83_000.png
```

Whereas the following seems to work

```
$ cd ~/shots
$ ls -la
$ scrot '%Y-%m-%d-%H%M%S_$wx$h.png'
$ ls -la
.rw-r--r-- 166k aureooms 14 Sep 11:34 2019-09-14-113409_1440x900.png
```
Seems reproducible.

daltomi:
> I am observing a weirder behavior on my machine.
> 
> ```
> $ pacman -Ss scrot
> community/scrot 1.2-1 [installed]
>     Simple command-line screenshot utility for X
> ```
> 
> The following creates two files
> 
> ```
[...]

PR #32, opened 2019-08-19
By default allow to overwrite a file.

daltomi:
Hi,
A possible solution to #30 

Now the `--overwrite` option has the default value of 1.

To **avoid overwriting** a file we would use: `scrot  shot.png --overwrite 0`
This create the `shot_000.png` file if it already exists.

Note, I don't use another option name because I can't think of another one and I'm lazy.

Chreers.

### DIFF
--- a/man/scrot.1
+++ b/man/scrot.1
@@ -105,8 +105,8 @@ Prevent beeping.
 Capture the mouse pointer.
 .TP
 .B
-\fB-o\fP, \fB--overwrite\fP
-By default \fBscrot\fP does not overwrite the files, use this option to allow it.
+\fB-o\fP, \fB--overwrite\fP NUM
+By default \fBscrot\fP overwrites the files, assign the value 0 to NUM to not overwrite.
 .TP
 .B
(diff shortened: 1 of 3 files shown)


GrayHatter:
I'm actually not in favor of making the default to allow overwriting... That would kinda defeat the intent of the setting, to protect dataloss.

I would like scrot to exit(non-zero) when it can't overwrite. I'm happy to open a pull to this effect if there's interest?

daltomi:
Hi,
> I would like scrot to exit(non-zero) when it can't overwrite. I'm happy to open a pull to this effect if there's interest?

From a keyboard shortcut, that would be a silent exit. 
What happens if you screenshot  is something really important, which happens sometimes only, a bug in a game for example?
I prefer the "always try to capture screen" behavior.

> I'm actually not in favor of making the default to allow overwriting... That would kinda defeat the intent of the setting, to protect dataloss.

Exactly, I also agree.

This PR I did more like a "revert", but the community did not participate asking for change, so now I close it.

PR #33, opened 2019-08-24
Resize the selection with the arrow keys.

daltomi:
Hi,
This PR use arrow keys to resize the start of the selection rectangle.
It is useful for correcting the selection.
@eribertomota if you agree you can merge it.
Cheers.


daltomi:
I will close it and try to join all these commits into one.
I also have to update it if the dependency to giblib is removed.

PR #40, opened 2020-03-01
Fix issue#39

### LINKED ISSUE #39: [Bug] Pointer in autoselect area not captured
When you pass the -a option and coordinates, even when the -p option is provided and the mouse pointer is in the area, it is not captured.

Hinigatsu:
2020-02-29-210103_1920x1080_scrot
This fix #39  
I used `scrot_grab_focused()` as an base. 

The screenshot shows the command used (`scrot -a 0,0,1920,1080 -p`) to test, and as shown, the mouse appears.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -64,7 +64,7 @@ main(int argc,
   else if (opt.select)
     image = scrot_sel_and_grab_image();
   else if (opt.autoselect)
-    image = gib_imlib_create_image_from_drawable(root, 0, opt.autoselect_x, opt.autoselect_y, opt.autoselect_w, opt.autoselect_h, 1);
+    image = scrot_grab_autoselect();
   else
   {
     scrot_do_delay();
@@ -525,6 +525,20 @@ scrot_sel_and_grab_image(void)
     if (opt.pointer == 1)
        scrot_grab_mouse_pointer(im, rx, ry);
   }
+  return im;
+}
+
+Imlib_Image
+scrot_grab_autoselect(void)
+{
+  Imlib_Image im = NULL;
+  int rx = opt.autoselect_x, ry = opt.autoselect_y, rw = opt.autoselect_w, rh = opt.autoselect_h;
+  Window target = None;
+
+  scrot_nice_clip(&rx, &ry, &rw, &rh);
+  im = gib_imlib_create_image_from_drawable(root, 0, rx, ry, rw, rh, 1);
+  if (opt.pointer == 1)
+	  scrot_grab_mouse_pointer(im, rx, ry);
   return im;
 }
 
--- a/src/main.c
+++ b/src/main.c
@@ -533,8 +533,8 @@ scrot_grab_autoselect(void)
 {
   Imlib_Image im = NULL;
   int rx = opt.autoselect_x, ry = opt.autoselect_y, rw = opt.autoselect_w, rh = opt.autoselect_h;
-  Window target = None;
 
+  scrot_do_delay();
   scrot_nice_clip(&rx, &ry, &rw, &rh);
   im = gib_imlib_create_image_from_drawable(root, 0, rx, ry, rw, rh, 1);
   if (opt.pointer == 1)
--- a/src/scrot.h
+++ b/src/scrot.h
@@ -81,6 +81,7 @@ void scrot_exec_app(Imlib_Image image, struct tm *tm,
 void scrot_do_delay(void);
 Imlib_Image scrot_sel_and_grab_image(void);
 Imlib_Image scrot_grab_focused(void);
+Imlib_Image scrot_grab_autoselect(void);
 void scrot_sel_area(int *x, int *y, int *w, int *h);
 void scrot_nice_clip(int *rx, int *ry, int *rw, int *rh);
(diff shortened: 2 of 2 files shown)


src/main.c
```diff
@@ -528,6 +528,20 @@ scrot_sel_and_grab_image(void)
   return im;
 }
 
+Imlib_Image
+scrot_grab_autoselect(void)
+{
+  Imlib_Image im = NULL;
+  int rx = opt.autoselect_x, ry = opt.autoselect_y, rw = opt.autoselect_w, rh = opt.autoselect_h;
+
+  scrot_do_delay();
```
Hinigatsu: I also took the liberty to add the delay option (`-d, --delay NUM`) to the autoselect function.

src/main.c
```diff
@@ -528,6 +528,20 @@ scrot_sel_and_grab_image(void)
...
+  scrot_nice_clip(&rx, &ry, &rw, &rh);
```
Hinigatsu: This `scrot_nice_clip()` is an already used function, but wasn't applied to the autoselect option.

Since this chances the autoselect option behavior (by adding an image size limit related to the screen size), I'm creating this this topic to know if this is a welcome feature.

src/main.c
ifohancroft: I am not sure if having the delay function auto applied is a good idea.

src/main.c
ifohancroft: Is the size limited to the size of a single screen or how does it work? What I mean is that I have two screens at the moment. Will I be able to take a screen shot from the second half of the left screen to lets say the end of the second screen? Essentially, a screen and a half size of an image.

Hinigatsu: ` scrot_nice_clip()` simply remove the "non-screen" space of the area when taking the screenshot.

I used `scrot_grab_focused()` function as a base, witch is called when the `-s` arg is passed. This means that the `scrot_grab_autoselect()` (`-a` arg) will mimic it.

I actually don't know how it affects a multi-monitor setup. I don't know if there's a practical way of emulating a second display, so could someone please clone and compile my branch and test this?! With and without the `-m` argument, and forcing a dimension that will grab both monitors!  

If this affects practicality, I then can a) think a better way of clipping the screen or b) simply remove this behavior!

Thanks for the attention on this detail! It has gone completely unnoticed :D

src/main.c
Hinigatsu: The `scrot_do_delay()` function is called, but only actively affects the flow when the `-d` arg is passed.

Since it was a missing option with the autoselect flow, I added in case of someone wanting to use!

ifohancroft: Ah! Yeah, that makes sense.

src/main.c
ifohancroft: I'll try to test it today, if not, tomorrow at the latest. Will also test the -m option but it has never really done anything for me (I mean I get the same result as I do without it. Maybe it works when you have multiple X screens?)

ifohancroft:
Sorry, I totally forgot to test. I will test it after dinner today.

src/main.c
ifohancroft: Sorry for the delay! I completely forgot to test the PR.

Here is what I found:
1. Making a screenshot of just one of monitors (done by pressing a key combo that executes a bash script that has the following line `scrot -a 0,0,1920,1080 -p -q 100 ~/Pictures/Screenshots/%s.png` for my first monitor and `scrot -a 1920,0,1920,1080 -p -q 100 ~/Pictures/Screenshots/%s.png`) works perfectly. The cursor is there and everything else is just like usual.

2. Making a screenshot of both of my monitors together (by executing `scrot -a 0,0,3840,1080` or `scrot -a 0,0,3840,1080 -m`) doesn't capture the cursor. Everything else seems to be just like usual.

3. Calling just `scrot` or `scrot -m` doesn't capture the cursor (I don't think it did before either) and like usual the screenshot between scrot and scrot -m looks the same (both monitors in the screenshot) and everything else also seems to be like usual.

I have noticed one weird thing but it may be usually like this:
When I type and execute `scrot -m` in terminal, later, in the screenshot, I see just `scrot -`
in the terminal. As if scrot somehow captured a screenshot back in time.

4. Here is where I have noticed a bug I think:
If I execute `scrot -a 0,0,3840,1080` or `scrot -a 0,0,3840,1080 -m` (since I have two 1920x1080 monitors configured next to each other) everything is still the same and like usual (even the 'm' is captured in the screenshot), however, the cursor is not being captured.
Maybe there is a certain area that if captured it will work, but generally, my tests have shown that if the area is made to be even a pixel bigger than the monitor but not enough to capture all of the monitors, the cursor won't show up. Not sure on which monitor's size does it depend though, as I don't have two differently sized monitors.

Hinigatsu: **Yay! It works!**
I'm glad to see that I've made this properly!
Thank you so much for testing.

PS: I couldn't replicate the `-m` behavior that you described. But since this PR is about the *autoselect* argument, I think we can revisit it in the future :D

ifohancroft: Most welcome!
Thanks so much for fixing it!

PS: I am also having trouble replicating the `-m` behavior but I don't think the problem with it is a result from this PR anyway.

PR #43, opened 2020-05-03
Fix required number. Fix autoselect.

spycapitan:
- options_parse_required_number(): requires that parameter is not null.

- options_parse_autoselect(): validate parameter format.


eribertomota:
Please, merge when done. I will send a invite for you now. Thanks!

PR #48, opened 2020-07-02
Fix SIGSEGV, option --note

spycapitan:
The functions next_space() and next_not_space() must detect end of C string
character.


eribertomota:
Hi @spycapitan!

Please, let me to know when we can release a new version.

Cheers,

Eriberto

spycapitan:
Hi @eribertomota 

For now I don't have any other fixes or features(maybe spycapitan#1), so it's a good opportunity to do it now.

Cheers.

PR #49, opened 2020-07-10
Fix SIGSEGV, if running without DISPLAY

ideal:
Currently if running on a server without DISPLAY set, it will cause segment fault.


ideal:
But how about running like this?

`DISPLAY=:1 scrot`

spycapitan:
Yes!, in that case you are right.
Just note that the `--display` parameter takes precedence over the `DISPLAY` environment variable.

PR #54, opened 2020-10-19
autotools: optionally debugging symbols

daltomi:
Currently scrot compiles with the debugging symbols which is no use to the end user,
it just ends up making the binary bigger.

My PC, x64:
```
With debug symbols    : 126K
Without debug symbols : 47K
```
The debugging symbols and the DEBUG definition is only useful for project collaborators or
anyone who requires it when opening an issue.

Another benefit between Debug and Release is that a contributor can place precondition code(assert) or use:

```C
#ifdef DEBUG
<print a value in the terminal only for the development stage>
#endif
```
Without having to worry that afterwards those code segments are in scrot, which would be useless to the end user.
It is also a communication tool between collaborators, informing others of the objective of a function or code segment.

To generate debug symbols:
```bash
./configure --enable-debug
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -20,6 +20,19 @@ AC_PATH_GENERIC(giblib, 1.2.3, [
   AC_SUBST(GIBLIB_CFLAGS) ],
   AC_MSG_ERROR(Cannot find giblib: Is giblib-config in the path?) )
 
+AC_ARG_ENABLE(debug,
+    [AS_HELP_STRING([--enable-debug],
+        [enable debugging code (default is no)])],
+    [debug="yes"], [debug="no"])
+
+if test "x$debug" = "xno"
+then
+    CFLAGS="-O3 -DNDEBUG"
+    LDFLAGS="-Wl,-s"
+else
+    CFLAGS="-g -O2 -DDEBUG"
+fi
+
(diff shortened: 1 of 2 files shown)


daltomi:
I've been looking at other projects that say `--disable-debug` or `--enable-release` and it's always the same, the `-g` flag is indicated at compile time.
The macro `M4` indicates this, unless the user manually assigns `CFLAGS`.
```
/usr/share/autoconf/autoconf/c.m4  
CFLAGS="-g"
```
The solution for the moment is to set the values to `CFLAGS`, `CPPFLAGS`, `LDFLAGS`, both when it is going to compile in debug or release mode.

|   MODE | CFLAGS | CPPFLAGS | LDFLAGS |
|---------|------------|----------|-------|
|  Release | -O3  | -DNDEBUG | -Wl,-s|
| Debug |  -O2 -g(db,db2) -Wall | -DDEBUG |   |

The most important, in release mode, is to define `-DNDEBUG` which disables assertion, and in debug mode is to define `-DDEBUG`.

daltomi:
Interesting:
https://www.gnu.org/software/autoconf-archive/ax_check_enable_debug.html

PR #59, opened 2020-11-05
Easy install

mrconter1:
Made the install easier. If 100 people have installed this repo they will have saved a total of 5 hours.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,14 @@ Nowadays, scrot is maintained by volunteers.
 
 ## Build and Install ##
 
+To build and install, run the following commands:
+
+    $ ./requirements
+    $ ./autogen.sh
+    $ ./configure
+    $ make
+    # make install
+    
 scrot depends of the following libraries/artifacts to build:
 
   - autoconf-archive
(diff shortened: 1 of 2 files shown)


jda:
This makes the build instructions specific to deb-style distros. Scrot is used on other distro flavors, and even on non-Linux systems such as OpenBSD. This would break for them.

If you want to assume that people don't have libs installed and want to provide, maybe include the deps in the build instructions in a subsection containing distro-specific steps. 

For example, you could drop the shell script and add the following after the build dep list in `README.md`:

> On Debian, you can install build dependencies with: `# apt-get install autoconf-archive giblib-dev libimage-imlib2-perl libtool libxcomposite-dev libxfixes-dev`

It's a little odd to specify libtool, but not build-essential. Also, rather rude to mutate whole-system state with `apt-get -y` hidden in a script.

Technical nits:
1. `./requirements` added to README, but file is called `requirements.sh`
2. `requirements.sh` is missing shebang, so you can't `./` it anyway.
3. Requirements script run as unprivileged user (`$` shell prompt, but no sudo before it or in requirements).

PR #64, opened 2021-01-23
Fix compilation on Sparc64 (on OpenBSD)

c0dev0id:
Hello,

I tried to compile scrot on OpenBSD/Sparc64 and it ran into compile errors due to two inline declarations that aren't strictly c99.

```
main.c: In function 'scrot_grab_mouse_pointer':
main.c:257: error: 'for' loop initial declaration used outside C99 mode
main.c: In function 'scrot_grab_stack_windows':
main.c:850: error: 'for' loop initial declaration used outside C99 mode
```

This pull request moves the declaration out of the for loops.

This fixes the compilation on sparc64 without relaxing the C standard.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -311,8 +311,9 @@ scrot_grab_mouse_pointer(const Imlib_Image image,
   pixels = (DATA32*)xcim->pixels;
 #else
   DATA32 data[width * height * 4];
-
-  for (size_t i = 0; i < (width * height); i++)
+  
+  size_t i;
+  for (i = 0; i < (width * height); i++)
     ((DATA32*)data)[i] = (DATA32)xcim->pixels[i];
 
   pixels = data;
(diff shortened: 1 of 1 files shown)


daltomi:
Hi,
I understand, that the statement in the `for` is allowed in C99.
The error message that your compiler indicates (I understand it must be GCC) [was corrected here.](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=31673)
What version of the compiler are you using?
BTW, I plan to accept this PR, as it is a minor change.

c0dev0id:
OpenBSD base comes with a pretty old gcc 4.2.1 (which is complemented by current clang and gcc in ports).
I can switch it to another compiler, but it's generally nicer when (especially small) tools don't depend on a big compiler as dependency.
Thanks for accepting this PR.

c0dev0id:
Also, I've patched this change into the port - so there's no pressure for a version update. But it's nice to keep ports and upstream in sync.

PR #78, opened 2021-03-27
Increase scrot_nice_clip() 'x' and 'y' by 1 when using --select

### LINKED ISSUE #77: unable to select 1px at the right size of the screen
image
look at the border of my bar. it's missing border at the right side because I am unable to select that area with scrot -s.

Hinigatsu:
fix #77


Hinigatsu:
After creating this PR I'm thinking that the "increase" should occur [inside scrot_sel_and_grab_image()](https://github.com/resurrecting-open-source-projects/scrot/blob/1bb7e509df6ec1bf281db593aaac823227f7ac80/src/main.c#L556)...

PR #79, opened 2021-03-27
Increase 'rx' and 'ry' size by 1 to avoid bad clipping when using --select

### LINKED ISSUE #77: unable to select 1px at the right size of the screen

Hinigatsu:
fix #77

I set the i3wm border to 1px and now scrot is able to grab it (the thin blue line at bottom and right edges)

image

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -555,6 +555,9 @@ key_abort_shot:
     }
     scrot_nice_clip(&rx, &ry, &rw, &rh);
 
+    ++rx;
+    ++ry;
+
     if (! opt.silent) XBell(disp, 0);
     im = gib_imlib_create_image_from_drawable(root, 0, rx, ry, rw, rh, 1);
 


daltomi:
Please see #80 , I think it is a better solution.

Hinigatsu:
@daltomi Thanks for taking a look! <3

One thing we have to consider is that `scrot_sel_and_grab_image()` is **always** grabbing -1 pixel than the actual selection rectangle at the bottom and right, but #80 only increases the clipping size when the selection is on the egde of the screen.

Maybe there's a better solution than mine, but we need to assert if this -1 pixel selection is acceptable.

daltomi:
Hi @Hinigatsu 
I was testing and I could not see what you indicate.
Anyway when I have more time I will continue testing. 
The #80  is maintained because it selects the last pixel on the screen. 

(Sorry for the overflow of images)
**User view:**

**Scrot capture:**


-------

**User view:**

**Scrot capture:**

Hinigatsu:
Strange!

The pixels under the left side of the rectangle are being grabbed:

(Note the rectangle selection line changing color where it align with the border of the window. You may have to zoom to see better details)

**User view:**
image

**Scrot capture:**
image

___

But the pixels under the right side aren't being grabbed:

**User view:**
image

**Scrot capture:**
image

daltomi:
I see. You are right.
Unfortunately I'm going on a trip in a few hours and I don't have time to collaborate.
See you in a few days. :)

Cheers.

Hinigatsu:
Have a nice trip! :D

daltomi:
Hi @Hinigatsu 
There I sent a new PR #81 , it takes into account the line width property, we are going to work with that one and close this one.
At the moment it works when the line width is 1px, but not when it is greater.

Cheers.

PR #84, opened 2021-05-28
Added 3 other corner cursors when selecting an area

Cationiz3r:
When **selecting** an area to screenshot from the **bottom left** corner to the **top right** corner, I noticed the cursor icon was a **south-east** angle one, which was kind of funny-looking, so I downloaded the source code and add more variables for cursor icons and conditions to display which.

I've also recompiled and installed the modified package on my end, and it worked out great!

### DIFF
--- a/src/scrot_selection.c
+++ b/src/scrot_selection.c
@@ -86,7 +86,10 @@ void scrot_selection_create(void)
     assert(sel != NULL);
 
     sel->cur_cross = XCreateFontCursor(disp, XC_cross);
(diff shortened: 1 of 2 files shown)


daltomi:
Hi @Cationiz3r 
I really like the effect it gives to the selection.

After merging your commit I will do some modifications and refactorings.
- Add the copyright to your name.
- Create 2 new functions to create and release the cursors.
- Make a single call to the XChangeActivePointerGrab function.

Thanks for contributing.

PR #86, opened 2021-06-03
Decouple giblib from scrot: resurrect/retire giblib as well? #22

languidnights:
giblib usage in scrot was primarily for simple wrappers around
imlib by setting context then calling/returning the wrapped
imlib_image_* calls. by reworking these calls, we can remove
the giblib dependency.

configure.ac: check for strdup
utils: implement strdup
slist: simple, dump singly-linked list
Makefile: add utils&slist
scrot: header updates
options: eprintf and strdup
main: remove simple wrappers with context + imlib calls

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -15,12 +15,17 @@ AC_PROG_MAKE_SET
 
 AM_MAINTAINER_MODE
 
-AC_PATH_GENERIC(giblib, 1.2.3, [
-  AC_SUBST(GIBLIB_LIBS)
-  AC_SUBST(GIBLIB_CFLAGS) ],
-  AC_MSG_ERROR(Cannot find giblib: Is giblib-config in the path?) )
+AC_PATH_GENERIC(Imlib2, , [
+  AC_SUBST(IMLIB2_LIBS)
+  AC_SUBST(IMLIB2_CFLAGS) ],
+  AC_MSG_ERROR(Cannot find Imlib2: Is imlib2-config in the path?) )
 
-LIBS="$LIBS -lm -lImlib2"
+
+LIBS="$LIBS -lm"
+IMLIB2_LIBS=`imlib2-config --libs`
+IMLIB2_CFLAGS=`imlib2-config --cflags`
+AC_SUBST(IMLIB2_LIBS)
+AC_SUBST(IMLIB2_CFLAGS)
 
 AC_CHECK_FUNC(getopt_long,,[LIBOBJS="$LIBOBJS getopt.o getopt1.o"])
 AC_SUBST(LIBOBJS)
--- a/configure.ac
+++ b/configure.ac
@@ -29,7 +29,7 @@ AC_SUBST(IMLIB2_CFLAGS)
 
 AC_CHECK_FUNC(getopt_long,,[LIBOBJS="$LIBOBJS getopt.o getopt1.o"])
 AC_SUBST(LIBOBJS)
-AC_CHECK_FUNC(strdup,,)
+AC_CHECK_FUNCS(strdup)
 
 m4_pattern_forbid([^AX_],[=> GNU autoconf-archive not present. <=])
 
--- a/configure.ac
+++ b/configure.ac
@@ -20,11 +20,7 @@ AC_PATH_GENERIC(giblib, 1.2.3, [
   AC_SUBST(GIBLIB_CFLAGS) ],
   AC_MSG_ERROR(Cannot find giblib: Is giblib-config in the path?) )
 
-LIBS="$LIBS -lm"
-GIBLIB_LIBS=`giblib-config --libs`
-GIBLIB_CFLAGS=`giblib-config --cflags`
-AC_SUBST(GIBLIB_LIBS)
-AC_SUBST(GIBLIB_CFLAGS)
+LIBS="$LIBS -lm -lImlib2"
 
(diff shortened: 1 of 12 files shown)


daltomi:
Thank you! 
This is a good candidate to close #22 
Please be patient while it is reviewed by other contributors.

languidnights:
I missed removing the header from one spot, and from the configure.ac script

configure.ac
```diff
@@ -20,11 +20,7 @@ AC_PATH_GENERIC(giblib, 1.2.3, [
   AC_SUBST(GIBLIB_CFLAGS) ],
```
daltomi, line 20: This must be removed too, look at my [PR](https://github.com/languidnights/scrot/pull/1/) that I sent you, you can merge it or copy and paste the changes that I indicate there

src/main.c
```diff
@@ -164,11 +171,13 @@ main(int argc,
       theight = cheight * opt.thumb / 100;
     }
 
+    imlib_context_set_anti_alias(1);
     thumbnail =
-      gib_imlib_create_cropped_scaled_image(image, 0, 0, cwidth, cheight,
-                                            twidth, theight, 1);
-    if (thumbnail == NULL)
-      gib_eprintf("Unable to create scaled Image\n");
```
daltomi, line 171: There are some replacements for `gib_eprintf` that don't exactly match its function.

```C
gib_eprintf: print error message and exit
gib_weprintf: print warning message and continue
```

configure.ac
languidnights: Thank you; I merged the PR

src/main.c
languidnights: Thank you; I added exits to the two places I removed eprintf without adding the exit

src/utils.h
```diff
@@ -0,0 +1,36 @@
...
+
+*/
+
+#ifndef SCROT_UTILS_H
+#define SCROT_UTILS_H
+
+#include <string.h>
+
+#ifndef HAVE_STRDUP
+#define strdup(a) _strdup(a)
```
daltomi, line 32: Hi @languidnights 
Some changes are needed: 
1) use the correct macro:`AC_CHECK_FUNCS` and `SCROT_HAVE_STRDUP`
2) prevent the function `_strdup` from being included in the binary if `#define SCROT_HAVE_STRDUP 1`:
```bash
nm src/scrot  | grep strdup
0000000000007900 T _strdup                                                                                                                                                                        
                 U strdup@GLIBC_2.2.5 
```
  what we want is a result like this, as I already said if `#define SCROT_HAVE_STRDUP 1`
```bash
nm src/scrot  | grep strdup
                 U strdup@GLIBC_2.2.5
```


 3) in case of failure `malloc` call to `exit(EXIT_FAILURE)` (giblib do that)

I will send a PR to your branch with the changes

src/slist.c
```diff
@@ -0,0 +1,62 @@
...
+
+  while(list->next) {
+    list = list->next;
+  }
+
+  return list;
+}
+
+int is_scrot_imlib_list_empty(Scrot_Imlib_List *list) {
+  if (list == NULL && list->next == NULL)
```
daltomi, line 58: Here exists a  possible null pointer dereference,  please correct that thanks.

languidnights: Thank you! a null dereference is guaranteed there if the list is null, not just possible. I removed the second condition, as it's not an empty list if list != null anway

src/options.c
```diff
@@ -433,8 +437,12 @@ options_parse_display(char *optarg)
...
      length = 256;
    }
-   new_display = gib_emalloc(length);
-   strncpy(new_display, optarg, length);
+   new_display = malloc(length);
+   if (! new_display) {
+     fprintf(stderr, "Unable to allocate display: %s", strerror(errno));
+     exit(EXIT_FAILURE);
+   }
+   strcpy(new_display, optarg);
```
daltomi, line 445: Please revert this to the function `strncpy`
The var. `length` has a maximum of 256 and `optarg` could be greater overlapping the memory hosted by `malloc`

src/utils.c
```diff
@@ -0,0 +1,67 @@
...
+Copyright 2019      Jade Auer <jade@trashwitch.dev>
+Copyright 2020      blockparole
+Copyright 2020      Cungsten Tarbide <ctarbide@tuta.io>
+Copyright 2020      daltomi <daltomi@disroot.org>
+Copyright 2020      Hinigatsu <hinigatsu@protonmail.com>
+Copyright 2020      nothub
+Copyright 2020      Sean Brennan <zettix1@gmail.com>
+Copyright 2020      spycapitan <spycapitan@protonmail.com>
+Copyright 2021      c0dev0id <sh+github@codevoid.de>
+Copyright 2021      Christopher Nelson <christopher.nelson@languidnights.com>
```
daltomi, line 19: The notice of `copyright` should  like the one you indicate in `src/utils.h` file.

languidnights: Updated the 4 new files to have the correct file names and the single line of copyright (which is I believe what you were asking?)

src/options.c
languidnights: I reverted it

src/utils.c
daltomi: Yes, thanks you.

daltomi:
Hi @languidnights 
I'm done with the review. 
Tomorrow I am going to merge the PR, if you have any last minute changes let me know, thanks 😃 
I decided to do a _Squash and merge_ for these 12 commits with the following title and description:

PR #92, opened 2021-07-10
add the Meson build system

guijan:
This appears to depend on #90 to build successfully. I will test it further once it's all merged in.


eribertomota:
Closing. We don't intend to change the install system. However, thanks for your contribution.

PR #93, opened 2021-07-10
README.md: partial rewrite

guijan:
Hi, I've polished the README.md file. 
The most inconsequential change is the correct English spelling and structure, but there is more to this patch:

I've removed redundant statements. For instance, repeating that scrot stands for (SCReenshOT) is jarring. 
The original text  also states that imlib2 is used to save images, procceeds to mention that imlib2 is used to support multiple image formats, and then lists support for multiple image formats among its features. That's a 3-way repetition.

There was one incorrect statement. scrot doesn't seem to depend on libtool directly. I'm not very good with autotools, so if this is wrong we can fix it.

I've clarified the build instructions. There was no mention that `make distclean` doesn't fully clean up the project, and that to do that `./autogen.sh clean` is needed. 

English punctuation is not so obvious and it was misused, so I've fixed it. 
For instance, a colon either precedes a listing or follows a complete clause and its use didn't always follow that.
The parenteses and the dot inside the parentheses in the list of supported image formats was also wrong.

That isn't everything, but that's the gist of it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ scrot requires a few projects and libraries:
 - libXext [(can be found in X.Org)](https://gitlab.freedesktop.org/xorg/lib/libxext)
 - libxfixes [(can be found in X.Org)](https://gitlab.freedesktop.org/xorg/lib/libxfixes)
 
-From a unix command line, building is simple:
+From a Unix command line, building is simple:
 ```
 $ ./autogen.sh
 $ ./configure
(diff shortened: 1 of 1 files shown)


guijan:
Oh, I also forgot to mention you can view the result of this new readme in [my fork](https://github.com/guijan/scrot/blob/polish-readme/README.md)

eribertomota:
About './autogen.sh clean', it also can be used before the ./configure command, without a 'make distclean'.

guijan:
Is it okay if the phrasing is not exactly "On Debian systems you can use"? That is actually incorrect English since it's not a complete clause.

eribertomota:
> On Debian systems



> Is it okay if the phrasing is not exactly "On Debian systems you can use"? That is actually incorrect English since it's not a complete clause.

"On Debian systems" plus a complement.

eribertomota:
However, "Debian users can install scrot from their package manager:" is ok.

PR #94, opened 2021-07-12
Remove local implementation functions

daltomi:
Proposal: Unify commits e5a23d8 and 92869db by correcting merge conflicts and adding a change to the configure.ac file

Hi @guijan
Please, any changes or if you do not agree with this PR please let me know.

I will wait a day or two before merging, thanks and regards!


guijan:
getopt_long is a GNU extension, it's not part of POSIX. It would be best to change the error message in configure.ac to reflect that.

daltomi:
I see, thanks

guijan:
I think it's perfect now.

PR #98, opened 2021-07-13
remove debug.h

guijan:
The macro DEBUG is not used anywhere within the source and it's also currently commented out.


daltomi:
Correct.
 I don't see that any standard needs it, for example GNU Coding Standards, it doesn't indicate anything.
This file seems to be more of a WIP than anything else, it's not a bad idea but we don't need it.

PR #99, opened 2021-07-14
add the Cirrus CI system

guijan:
This automates build testing in the following platforms:
  - Alpine Linux (musl Linux)
  - Debian Linux (glibc Linux)
  - FreeBSD
  - MacOS

Multiple branches of Debian are used to cover toolchains ranging from ancient to bleeding edge.
It would be very easy to add testing to this in the future, but
currently scrot has no tests.

If this is merged in, @eribertomota will have to add the
[Cirrus CI](https://cirrus-ci.org/) application to his github account to enable it.

Cirrus can be seen in action at this URL: https://cirrus-ci.com/github/guijan/scrot/add-cirrus

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,112 @@
+# Copyright (C) 2021 Guilherme Janczak <guilherme.janczak@yandex.com>
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to
+# deal in the Software without restriction, including without limitation the
+# rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+# sell copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies of the Software and its documentation and acknowledgment shall be
+# given in the documentation and software packages that this Software was
+# used.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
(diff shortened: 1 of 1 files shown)


eribertomota:
IMO, it is a good idea! However, as a Debian Developer, I would like to warn about oldstable and stable. I think is not a good idea test against stable and, specially oldstable, because the libs are very old and it can generate cyclic building fails. Are you sure to keep stable and oldstable? In Debian, 99% of the packages are sent to unstable. Testing is very similar to unstable and is a good idea keep it.

@daltomi is ok for you add this CI system?

guijan:
I think it's worth it to keep oldstable and stable. 

They tell us exactly when we can drop deprecated features. For instance, the current Autotools setup uses some deprecated macros as the alpine-edge-clang build tells us:
```
configure.ac:6: warning: The macro `AC_CONFIG_HEADER' is obsolete.
configure.ac:6: You should run autoupdate.
./lib/autoconf/status.m4:719: AC_CONFIG_HEADER is expanded from...
configure.ac:6: the top level
configure.ac:10: warning: 'AM_PROG_CC_STDC': this macro is obsolete.
configure.ac:10: You should simply use the 'AC_PROG_CC' macro instead.
```

When the build on Debian oldstable issues a warning about them it'll be a safe time to replace them.

eribertomota:
Em qua., 14 de jul. de 2021 às 17:38, guijan ***@***.***>
escreveu:

> I think it's worth it to keep oldstable and stable.
>
> They tell us exactly when we can drop deprecated features. For instance,
> the current Autotools setup uses some deprecated macros as the
> alpine-edge-clang build tells us:
>
> configure.ac:6: warning: The macro `AC_CONFIG_HEADER' is obsolete.configure.ac:6: You should run autoupdate.
> ./lib/autoconf/status.m4:719: AC_CONFIG_HEADER is expanded from...configure.ac:6: the top levelconfigure.ac:10: warning: 'AM_PROG_CC_STDC': this macro is obsolete.configure.ac:10: You should simply use the 'AC_PROG_CC' macro instead.
>
> When the build on Debian oldstable issues a warning about them it'll be a
> safe time to replace them.
>

I have worked with packaging in Debian since 2006. Commonly, all projects
try to keep the source code updated to most recent systems. I never saw an
upstream (mainstream) developing a code, thinking in a very old system.
Debian has the oldstable release, but after a new stable to be released,
oldstable has support for one year only. Debian never will provide a new
upstream release for stable or oldstable. New releases are for unstable
only.

I will wait for @daltomi ***@***.***>

Regards,

Eriberto

daltomi:
Hi everyone,

> @daltomi is ok for you add this CI system?

@eribertomota , yes I like the idea too.

Issue #42, opened 2020-04-21
publish release tarballs (including configure)


stapelberg:
Thanks so much for maintaining scrot!

When downloading a release (e.g. https://github.com/resurrecting-open-source-projects/scrot/archive/1.3.tar.gz, the latest at the time of writing), the tarball does not contain a configure script, only a configure.ac.

While it is fine to require autoreconf during development, release tarballs are supposed to be created using `make dist` and contain an already-generated configure. This will make packaging easier for downstream distributions.

This is even more important now that you’re requiring autoconf-archive, which is not necessarily as widely available as autoconf/automake themselves.

Could you please use `make dist` for releases going forward? Thank you.

PR #103, opened 2021-07-24
delete scrot.spec

guijan:
This kind of file is what Fedora uses to build its packages. 
It's outdated, wrong (our license is MIT-advertising not MIT), incomplete ([check their instructions on how to make a package](https://docs.fedoraproject.org/ro/Fedora_Draft_Documentation/0.1/html/RPM_Guide/ch-specfiles.html), our scrot.spec doesn't even tell Fedora how to build the package), and whatever we do the Fedora maintainers will (and should) maintain their own .spec file in their repository.

Also, it seems we're not currently packaged on Fedora, they still run scrot 0.8 from the original developer.


daltomi:
Still I do not know about this, but if approved a change needs: https://github.com/resurrecting-open-source-projects/scrot/blob/c99b06ac475dd6760c2cad81ff663aeb18c83e2c/Makefile.am#L45

daltomi:
You seem to be right.
The git history shows that this file was removed here:  6cce85f ("Removed autogenerated files", 2019-02-12)
After a long time, I wanted to solve the #42 with 626f030 ("autotools: support make dist")
I don't remember correctly, but as you can see, the file is in `Makefile.am` and I create it forcibly :-/

PR #101, opened 2021-07-23
polish the txt2man manual

guijan:
I've made changes to the txt2man manual.

The most innocuous ones are the minor English corrections and making the description like the README.md file in the source tree.
I've changed the line length to be at most 80 characters always, 80 being the "standard" amount of columns in a terminal.

The options are sorted alphabetically and case insensitively by their short forms. The upper case short options don't all come first.
The synopsis is now a complete summary of all the flags.
Some flags which didn't previously list an argument but took one now have the fact that they take arguments clearly specified. I'm not 100% sure I got all of them though.
The manual now uses exclusively the short options while explaining its flags, to encourage their use.

This isn't all but that's the gist of it.

PR #104, opened 2021-07-24
use the manual's synopsis as help text

guijan:
The old help text is outdated and a duplicate of the manual. This patch brings the help flag in line with the style popular among Suckless and BSD utilities:
```
$ src/scrot -h
usage:  scrot [-bcfhkmopsuvz] [-a X,Y,W,H] [-C NAME] [-D DISPLAY] \
              [-d SEC] [-e CMD] [-l STYLE] [-n OPTS] [-q NUM] [-S CMD] \
              [-t NUM | GEOM] [FILE]
```

Here's an example of a suckless utility which follows the same style:
```
$ dmenu -h
usage: dmenu [-bfiv] [-l lines] [-p prompt] [-fn font] [-m monitor]
             [-nb color] [-nf color] [-sb color] [-sf color] [-w windowid]
```

And an OpenBSD utility with the same style:
```
$ signify -h
signify: unknown option -- h
usage:  signify -C [-q] [-p pubkey] [-t keytype] -x sigfile [file ...]
        signify -G [-n] [-c comment] -p pubkey -s seckey
        signify -S [-enz] [-x sigfile] -s seckey -m message
        signify -V [-eqz] [-p pubkey] [-t keytype] [-x sigfile] -m message
```

As you can see there are minor differences, signify uses 2 spaces between "usage:" and the command's name whereas dmenu uses 1, and I added an inverse slash before the newline to tell the user the next line is a continuation of the previous and dmenu didn't. There's no strict definition of this style.

This patch also removes the function show_mini_usage(), which does the
same thing and is dead code.

edit: brought the text in line with the current state of #101

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -13,6 +13,7 @@ Copyright 2019-2021 Daniel T. Borelli <daltomi@disroot.org>
 Copyright 2019      Jade Auer <jade@trashwitch.dev>
 Copyright 2020      Sean Brennan <zettix1@gmail.com>
 Copyright 2021      Christopher R. Nelson <christopher.nelson@languidnights.com>
+Copyright 2021      Guilherme Janczak <guilherme.janczak@yandex.com>
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to
@@ -212,7 +213,6 @@ static void options_parse_window_class_name(const char* window_class_name)
         opt.window_class_name = strdup(window_class_name);
     }
 }
-
 
 static void
 scrot_parse_option_array(int argc, char **argv)
@@ -508,7 +508,6 @@ int options_cmp_window_class_name(const char* target_class_name)
     return !!(!strncmp(target_class_name, opt.window_class_name, MAX_LEN_WINDOW_CLASS_NAME - 1));
 }
 
-
 void
 show_version(void)
 {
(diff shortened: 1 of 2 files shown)


daltomi:
I don't know about this yet.

Good:
- I like the idea that only would have to keep the `man page` and not duplicate efforts.
- It is avoided to load a "lot" of data into memory (almost a `man page`) just to capture an image.
- GNU Coding Std say: _In the GNU project, **man pages are secondary**. **It is not necessary or expected** for every
GNU program to have a man page, but some of them do. **It’s your choice** whether to include
a man page in your program._
So, if all systems when you install `scrot` install the `man page`, that's fine because for a user the summarized help will seem cryptic. I suppose that the systems seeing that the package has a `man page` will install it

Bad:
- Breaks the user's habit: in blogs, wikis and others, most indicate `scrot --help`

eribertomota:
From Debian: "Each binary in /usr/bin, /usr/sbin, /bin, /sbin or /usr/games should have a manual page. [...] If no manual page is available, this is considered as a bug and should be reported to the Debian Bug Tracking System".

More details here[1][2].

[1] https://lintian.debian.org/tags/no-manual-page
[2] https://www.debian.org/doc/debian-policy/ch-docs.html#manual-pages

daltomi:
I'm going on a trip again, I leave some doubts.

Does it look better in lowercase and without the `\` character?

```bash
usage:  scrot [-bcfhkmopsuvz] [-a x,y,w,h] [-C name] [-D display]
              [-d sec] [-e cmd] [-l style] [-n opts] [-q num] [-S cmd]
              [-t num | geom] [file]
```
Or capitals are intentional to match the `man page`?

Cheers!!

PR #111, opened 2021-07-27
add libbsd dependency

guijan:
This patch does a few things all at once:
- add libbsd to configure.ac
By default, the configure script picks up libbsd if it is necessary. A "--without-libbsd" ./configure option was added to allow package maintainers to test whether the package is needed on their OS without poking into the source or removing libbsd from their local installation.

- add the libbsd effort to TODO.md
I also deleted the plaintext TODO, which was nothing but a single line joke.

- add libbsd package to Cirrus CI and Github Actions
Currently, the libbsd package is only installed on Debian and Ubuntu among our tested systems because all the other systems provide all the currently used functions.
Because they were on the same line, I also removed the giblib and libtool packages from the Github Actions CI-scrot does not currently depend on either.

- add the first usage of each of err(), strlcat(), and strlcpy().
It turned out that the "err" identifier was in use in the code, so a variable had to be renamed.

- add libbsd to the dependency list at README.md

- point to TODO.md from CONTRIBUTING.md


daltomi:
Hi,
Unfortunately I have an error in GNU/Linux. I have the `libbsd` library installed.
(a quick test, I didn't see your commit in depth, sorry)

```bash
./configure --without-libbsd
...
checking for strlcpy... no
checking for strlcat... no
checking for err... yes
checking for errx... yes
configure: error: BSD functions not found and --without-libbsd was used
```
AFAIK, the `err/errx`  functions are from Linux:
`man 3 err - Linux Programmer's Manual`

```
CONFORMING TO
       These functions are nonstandard BSD extensions.
```
That is why `autotools` finds these functions but not the `strl[..]`
The ones that are from `libbsd` are `errc` and family.

guijan:
That is intentional. 
"./configure --without-libbsd" hard-fails if libbsd is required
"./configure" uses the system functions if they're found, and if any of them is missing the script will add libbsd as a dependency.

"--without-libbsd" exists so maintainers can test whether it's needed easily.

daltomi:
Ok I see.

src/main.c
```diff
@@ -174,10 +174,8 @@ main(int argc,
       imlib_create_cropped_scaled_image(0, 0, cwidth, cheight,
                                             twidth, theight);
     if (thumbnail == NULL) {
-      fprintf(stderr, "Unable to create scaled Image: %s\n", strerror(errno));
-      exit(EXIT_FAILURE);
-    }
-    else
+        err(EXIT_FAILURE, "imlib_create_cropped_scaled_image");
```
daltomi, line 177: We must avoid informing the user of issues internal to the code since it does not really inform him of the error. Please leave it as it was, but feel free to expand the description of the error if you want.

daltomi (changes requested):

src/main.c
daltomi: I made changes to your branch resolving conflicts, remember to update your local branch with: `git pull` before making changes.

configure.ac
```diff
@@ -22,6 +22,22 @@ PKG_CHECK_MODULES([XEXT], [xext])
...
 
+AC_ARG_WITH([libbsd],
+    AS_HELP_STRING([--without-libbsd], [Error when BSD functions are not found]))
+AC_CHECK_FUNCS([strlcpy strlcat err errx],, [LIBBSD_NEEDED=yes])
+AS_IF([test "x$LIBBSD_NEEDED" = "xyes"], [
+    AS_IF([test "x$with_libbsd" = "xno"], [
+        AC_MSG_ERROR([BSD functions not found and --without-libbsd was used])
+    ])
+    PKG_CHECK_MODULES([LIBBSD], [libbsd-overlay],,
+        AC_MSG_ERROR([BSD functions not found, libbsd is required]))
```
daltomi, line 33: This macro has a syntax error.
On systems where `libbsd` is not installed this macro generates an error and creates a junk file.
```bash
checking for strlcat... no
checking for err... yes
checking for errx... yes
checking for LIBBSD... no
configure: error: BSD functions not found
elif test yes = untried; then
        { printf %s\n configure:5389:
```
This error creates a new file called: `no >&5?printf %sn no >&6; }??as_fn_error 0 BSD`

```bash
 stat no\ \>\&5$'\n'printf\ %sn\ no\ \>\&6\;\ \}$'\n'$'\t'as_fn_error\ 0\ BSD
  File: 'no >&5'$'\n''printf %sn no >&6; }'$'\n\t''as_fn_error 0 BSD'
  Size: 116             Blocks: 24         IO Block: 4096   regular file
```
 
 Patch:
 ```patch
 
diff --git a/configure.ac b/configure.ac
index 769b8b7..563f5f8 100644
--- a/configure.ac
+++ b/configure.ac
@@ -30,7 +30,7 @@ AS_IF([test "x$LIBBSD_NEEDED" = "xyes"], [
         AC_MSG_ERROR([BSD functions not found and --without-libbsd was used])
     ])
     PKG_CHECK_MODULES([LIBBSD], [libbsd-overlay],,
-        AC_MSG_ERROR([BSD functions not found, libbsd is required]))
+        [AC_MSG_ERROR([BSD functions not found, libbsd is required])])
 ])
```

daltomi (changes requested):

daltomi (approved):

PR #115, opened 2021-07-30
The delay option should not be a negative number

daltomi:


guijan:
Hi, both this and #114 would be better served by [strtonum()](https://manpages.debian.org/unstable/libbsd-dev/strtonum.3bsd.en.html)

daltomi:
Okay, but not now, it is not so easy to use this range-based function.
Some code blocks use their own range check informing the user of their error.
For example:
```C
 opt.line_opacity = options_parse_required_number(value);

  if (opt.line_opacity < 10 || opt.line_opacity > 100) {
    // error Value of the range (10..100) 
   // exit
   }
```
We must find the way that the input from the user was an opacity range error and not an internal error to strtonum.

PR #119, opened 2021-07-30
add an extra entry to .gitignore

guijan:
I use a tool called [Bear](https://github.com/rizsotto/Bear) to generate a [compile_commands.json](https://clang.llvm.org/docs/JSONCompilationDatabase.html) file. It tells my Vim setup how to compile scrot and run linters automatically.

I also sorted the file while at it.


daltomi:
Ok, everything that helps collaborators is useful for the project.

PR #120, opened 2021-07-30
assume c89 functions exist

guijan:
This changes configure.ac to assume c89 functions exist.
Checking for them is pedantic and scrot is not compatible with C89 or pre-ANSI C, scrot is written in the newer C99 standard.

The more obtuse and roundabout the code is the harder it is to understand and maintain. Additionally, this will speed up the build a little as autotools is fairly slow.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -51,7 +51,7 @@ AC_CHECK_TYPES([ptrdiff_t])
 
 # Required: Checks for library functions.
 AC_FUNC_MALLOC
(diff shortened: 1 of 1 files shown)


guijan:
Apologies for being unclear. What I mean is that it's past the point of deprecation to check for whether functions introduced in a past standard are present. scrot is written in C99 which is actually still not fully implemented in the wild, but it's already time to stop worrying about whether C89 is fully implemented.

PR #122, opened 2021-07-31
remove obsolete autotools macros

guijan:
This patch:
- removes the the still hardcoded include directories in Makefile.am.
- updates the AC_CONFIGURE_HEADER macro to AC_CONFIGURE_HEADERS
Our CI says it's obsolete, I can't find the exact autotools announcement about this, but here's someone on a GNU website saying it's obsolete already in 2006:
https://savannah.gnu.org/support/?105403
- removes the obsolete AM_PROG_CC_STDC macro
Oddly, we were using this obsolete macro as well as its replacement (AC_PROG_CC) at the same time.
Look for AM_PROG_CC_STDC at https://autotools.io/forwardporting/automake.html
- adds the AC_PROG_CC_C99 macro
It sets the compiler to C99 mode if that isn't the default. This has the side effect of making the build work with ancient GCC versions in which C99 is supported and C89 is the default, e.g. GCC 4.2.
- removes the AC_PATH_X macro
We get X compiler flags from pkg-config now. Additionally, the X flags provided by this macro weren't actually used anyway.
- removes checks under the "checks for typedefs, structures, ..." comment
All of those except the stdbool one check for C89 features.
- removes the AC_FUNC_MALLOC macro
This macro is a workaround for malloc implementations that don't behave like the GNU one when a malloc(0) call is made (allocating 0 bytes is implementation-defined behavior in C). It's trivial to write correct C where this isn't a problem, and scrot has few malloc/calloc calls and already seems to do it.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -3,6 +3,7 @@
 #                                   scrot_sucks@linuxbrit.co.uk>
 # Copyright 2015-2019 Joao Eriberto Mota Filho <eriberto@eriberto.pro.br>
 # Copyright 2019-2020 Daniel T. Borelli <daltomi@disroot.org>
+# Copyright 2021      Guilherme Janczak <guilherme.janczak@yandex.com>
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to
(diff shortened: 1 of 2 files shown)


daltomi:
Okay, but aren't you contradicting what you said here?  https://github.com/resurrecting-open-source-projects/scrot/pull/99#issuecomment-880188349

If we merge we should follow the advice of @eribertomota

daltomi:
Hi @guijan 
It was merged into #136  with other required changes.

PR #123, opened 2021-07-31
CI: Use "apt-get"

keysym:
The [apt manpage](https://manpages.ubuntu.com/manpages/xenial/en/man8/apt.8.html) states that "apt-get" should be used in scripting.

### DIFF
--- a/.github/workflows/full-check.yml
+++ b/.github/workflows/full-check.yml
@@ -15,7 +15,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: install_dependencies
-      run: sudo apt install autoconf-archive libimlib2-dev libxcomposite-dev libxfixes-dev libbsd-dev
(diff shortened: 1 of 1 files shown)


eribertomota:
Hi @keysym,

Thanks for your contribution.

The man apt says:

_All features of apt(8) are available in dedicated APT tools like apt-get(8) and apt-cache(8) as well.  apt(8) just changes the default value of some options (see apt.conf(5) and specifically the Binary scope). So you should prefer using these commands (potentially with some additional options enabled) in your scripts as they keep backward compatibility as much as possible._

It is from Xenial, released in 2016. Currently, GitHub uses Ubuntu 20.04, that is full compatible with APT. I don't see here good enough reason to change from stable and well maintained in Debian "APT" to apt-get.  Independent of anything, the current script is working very fine in GitHub.

keysym:
The `apt` in the pipelines produces a warning:
https://github.com/resurrecting-open-source-projects/scrot/runs/3210017594#step:3:5

But yeah, using `apt-get` instead would not be a substantial change and is more about "best practices"!

Issue #38, opened 2020-02-22
[Feature request] Select monitor


ifohancroft:
Ability to select which monitor to take a screenshot of (the full monitor) on a multiple monitors setup.

ifohancroft:
Also, since executing 'scrot' works the same as 'scrot -m' on a multiple monitor setup, can we get the -m option to be for the select monitor? Also, the long option --multidisp can be dropped and it can be changed to --monitor

dirdi:
This is already scriptable. The following command takes a screenshot of (and only of) the first monitor:
```
scrot -a $(xrandr | grep '\Wconnected\W' | grep -Po "\d+x\d+\+\d+\+\d+" | sed 's/x/+/' | awk -F "+" '{print $3","$4","$1","$2}' | head -n 1 | tail -n 1)
```

ifohancroft:
Thanks but I know that. My feature request stands despite that.

eribertomota:
Em dom., 19 de abr. de 2020 às 17:46, IFo Hancroft <notifications@github.com>
escreveu:

> Thanks but I know that. My feature request stands despite that.
>

I agree. The right way is  add the feature in scrot. But thanks for this
workaround.

Now, we need new volunteers to add new features in scrot.

ifohancroft:
If someone gets to doing it, here's what I have:

The libxrandr library holds/provides the RandR extension. Package is libxrandr but depends on distro probably.

The following functions found there (mostly in Xrandr.h but in some of the other files as well) might help you:
XRRGetOutputPrimary
XRRGetMonitors
XRRGetScreenSizeRange

Keep in mind that you will probably need to check the configured screen arrangement (where they are located in relation to one another) as the order in which you get them, may not be the order they are set in). For example my HDMI display is on the left and set as primary, while my DVI display is on the right, however the xrandr binary, lists the DVI first (also it is the first one to boot and sits on the left before xrandr is loaded to read its configuration)

PR #137, opened 2021-08-15
New feature: select monitor (#38)

daltomi:
Ref: #38 
New option: `--monitor`, `-M`

- First attempt with `libxrandr`, temporarily discarded
- Second attempt with `libxinerama`, simpler API.

Example: `scrot -M 8` (force to fail, see log)


daltomi:
It is a draft. 
This is a very old branch but `autotools` is configured correctly.
Resolving all conflicts with the `master branch` will never be done.

daltomi:
To test `xrandr`:
```bash
git checkout -b xrandr 0a23c77
```

ifohancroft:
I'm very happy to see that someone is working on this!

What was your reason for discarding the libxrandr method btw?

daltomi:
It just seems to me that Xinerama is simpler to use. For now it works in the tests I did.

ifohancroft:
> It just seems to me that Xinerama is simpler to use. For now it works in the tests I did.

To be honest, I was suprised it works with Xinerama without the monitors being in Xinerama mode (where each monitor is on its own X server). Unless of course I am mixing up terms.

PR #142, opened 2021-08-19
New feature: hide/hole selection area (improvement)

daltomi:
Highlight the selected area overshadowing the rest of the capture.
The shadow color is the same as the line color.

Ref: 95bea6b (New feature: hide selection area)

selection-hole0-fs8
selection-hole1-fs8


eribertomota:
Very good @daltomi !

daltomi:
News: af132df (Added support for alpha color)

The `opacity` suboption of the line style is used for the alpha channel: `capture`(line alpha), `hole`(rectangle alpha), `hide`(rectangle alpha)

Previously if `opacity` was less than 10 or greater than 100, the program was terminated, now, we will indicate to the user, via man page, that the range for opacity is between 0 and 255.
The program **handles the ranges internally**, for example the selection mode `edge` has its own range(10-100)

The mechanism of the option `--select` to create an area via selection of a window by mouse click is maintained.

select
In both images I **clicked on the terminal window**, what changed was the action of `--select`, on the left we see `hide` and on the right `hole`:
```bash
scrot --pointer --select=hole --line color="red",opacity=140
```
Remember: option `--select` without suboption implies: `--select=capture`,  default `opacity` 100

TODO: (I'm going on 🚙 , yeah again!)
  - [ ]  Update manpage

daltomi:
> Very good @daltomi !

Thanks 😃 
FYI, now I use Devuan(🩹 Debian) :P

eribertomota:
> > Very good @daltomi !
> 
> Thanks smiley
> FYI, now I use Devuan(adhesive_bandage Debian) :P

Very good again!

PR #147, opened 2021-08-27
Support redirect to stdout

daltomi:
Allows the form:
```bash
$ scrot - > file.png

$ scrot -> file.png

$ scrot - | pngquant -o file.png - 
etc..
```

The only thing this PR does is try the following in a transparent way: `scrot -o /dev/stdout`
"Transparently" means that not every system has a `/dev/stdout`.
If no `/dev/stdout` is found it tries `/dev/fd/`

Please, if anyone knows another way feel free to tell me, thank you.

If you want to test how it prints in your terminal, I advise you to use the `--select` option and select a very small area.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -223,25 +223,30 @@ static void optionsParseWindowClassName(const char* windowClassName)
         opt.windowClassName = strndup(windowClassName, MAX_LEN_WINDOW_CLASS_NAME);
 }
 
-static bool accessFileOk(char* pathName)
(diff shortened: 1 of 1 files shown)


daltomi:
TODO:

- [ ] man page (clarify that the image format is only in PNG by default which is the format for files with names without extension)

daltomi:
Tested on NetBSD, ok: `/dev/stdout`

PR #152, opened 2021-10-03
New option --file (-F): Adds the ability to provide the filename as an option

ifohancroft:
It adds the ability to provide the filename as --file (-F) instead of having to necessarily have it at the end.
(This doesn't remove the ability to just pass the filename at the end or to just execute scrot without providing a filename at all.)
Makes it easier to alias if you always save your files in the same directory, as well as easier to use in scripts.


ChangeLog
```diff
@@ -1,3 +1,10 @@
+Sun Oct 03 01:50:00 -03:00 2021  IFo Hancroft <contact@ifohancroft.com>
```
daltomi, line 1: We do not change this file until the release of a new version, which has not been discussed yet and the one in charge of this is @eribertomota 
However, the version number that you changed indicates a patch, which is not.

man/scrot.txt
```diff
@@ -3,7 +3,7 @@ NAME
 
 SYNOPSIS
   scrot [-bcfhkmopsuvz] [-a X,Y,W,H] [-C NAME] [-D DISPLAY] [-d SEC] [-e CMD]
-        [-l STYLE] [-n OPTS] [-q NUM] [-S CMD] [-t NUM | GEOM] [FILE]
+        [-f FILE] [-l STYLE] [-n OPTS] [-q NUM] [-S CMD] [-t NUM | GEOM] [FILE]
```
daltomi, line 6: It must be capitalized: `-F`

src/options.c
```diff
@@ -433,6 +431,18 @@ void optionsParseThumbnail(char* optarg)
...
 
+void optionsParseFileName(char* fileName)
+{
+    opt.outputFile = fileName;
+
+    if (strlen(opt.outputFile) > MAX_OUTPUT_FILENAME)
+        errx(EXIT_FAILURE,"output filename too long, must be "
+            "less than %d characters", MAX_OUTPUT_FILENAME);
+
+    if (opt.thumb)
```
daltomi, line 442: This only worked (note the past) when the filename was the last parameter.

This way it will fail: `scrot -F file.png -t 50%`

Because the variable `opt.thumbFile`  will be unassigned.
The next patch could fix the problem. If you agree, implement it yourself.

```patch
diff --git a/src/options.c b/src/options.c
index c0feada..6426567 100644
--- a/src/options.c
+++ b/src/options.c
@@ -339,6 +339,9 @@ void optionsParse(int argc, char** argv)
             warnx("unrecognised option %s", argv[optind++]);
     }
 
+    if (opt.thumb && opt.outputFile)
+        opt.thumbFile = nameThumbnail(opt.outputFile);
+
     /* So that we can safely be called again */
     optind = 1;
 }
@@ -438,9 +441,6 @@ void optionsParseFileName(char* fileName)
     if (strlen(opt.outputFile) > MAX_OUTPUT_FILENAME)
         errx(EXIT_FAILURE,"output filename too long, must be "
             "less than %d characters", MAX_OUTPUT_FILENAME);
-
-    if (opt.thumb)
-        opt.thumbFile = nameThumbnail(opt.outputFile);
 }
 
 void optionsParseNote(char* optarg)
``

File: ChangeLog (hunk 1 above)
Comment (ifohancroft) [reply ChangeLog:1]: I have not changed the existing functionality. I have added an extra option.

As for the version number - what is the appropriate version number?

File: src/options.c [hunk 4]
```diff
@@ -469,7 +479,7 @@ void showUsage(void)
     fputs(/* Check that everything lines up after any changes. */
         "usage:  " SCROT_PACKAGE " [-bcfhkmopsuvz] [-a X,Y,W,H] [-C NAME] [-D DISPLAY]"
         "\n"
-        "              [-d SEC] [-e CMD] [-l STYLE] [-n OPTS] [-q NUM] [-S CMD] \n"
+        "              [-f FILE] [-d SEC] [-e CMD] [-l STYLE] [-n OPTS] [-q NUM] [-S CMD] \n"
```
daltomi, line 482: It must be capitalized: `-F`

daltomi (changes requested):

ChangeLog
daltomi: See: https://semver.org/

man/scrot.txt
ifohancroft: Thanks! I'll fix that. It's a typo, I meant to capitalize it.

src/options.c
ifohancroft: Thank you! I'd look into this.

src/options.c
ifohancroft: Ugh... I can't believe I made the same typo in two different files..

ChangeLog
ifohancroft: Thank you!

Should I remove what I added from ChangeLog or should I change it to 1.7 (since that's the correct way according to semver for new features with backwards compatiblity)

ifohancroft:
I made the requested changes. Let me know if I missed or perhaps misunderstood something or if there is anything more you'd like me to change.

ChangeLog
daltomi: We never modify the files `ChangeLog` , `create-man.sh` and `scrot.1`, are files that are modified when launching a new release, that is in charge of @eribertomota 

You can modify `scrot.txt`

ifohancroft: Ah! Got it. Thank you!
I will revert the changes when I wake up later today.

ifohancroft: I think it should be ok now.

daltomi (approved):

PR #155, opened 2021-10-05
File option cleanup

ifohancroft:
When I initially implemented the --file / -F option, the naming of the thumbnail was done in the options.c file, at the end of the options parsing function, by introducing an extra check if there is a filename and a thumbnail. However, I have felt that there must be a better way / place to do that.

I have since found out that in main.c after the parsing of the options, there is already a check if a filename exists and naming of both the thumbnail and the filename is done there. This felt like the perfect place.

So, from options.c, I removed:

```
if (opt.thumb && opt.outputFile)
        opt.thumbFile = nameThumbnail(opt.outputFile);
```

In main.c I changed this:

```
if (!opt.outputFile) {
        opt.outputFile = strdup("%Y-%m-%d-%H%M%S_$wx$h_scrot.png");
        opt.thumbFile = strdup("%Y-%m-%d-%H%M%S_$wx$h_scrot-thumb.png");
    } else
        scrotHaveFileExtension(opt.outputFile, &haveExtension);
```

To this:

```
if (!opt.outputFile) {
        opt.outputFile = strdup("%Y-%m-%d-%H%M%S_$wx$h_scrot.png");
        opt.thumbFile = strdup("%Y-%m-%d-%H%M%S_$wx$h_scrot-thumb.png");
    } else {
        opt.thumbFile = nameThumbnail(opt.outputFile);
        scrotHaveFileExtension(opt.outputFile, &haveExtension);
    }
```

However, there turned out to be a bug inside of nameThumbnail, that was breaking it when the call to it was moved to main.c This also fixes that, as well as converts the function to using strlcpy and strlcat instead of strncpy and strlcat.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -84,8 +84,10 @@ int main(int argc, char** argv)
     if (!opt.outputFile) {
         opt.outputFile = strdup("%Y-%m-%d-%H%M%S_$wx$h_scrot.png");
         opt.thumbFile = strdup("%Y-%m-%d-%H%M%S_$wx$h_scrot-thumb.png");
-    } else
+    } else {
+        opt.thumbFile = nameThumbnail(opt.outputFile);
         scrotHaveFileExtension(opt.outputFile, &haveExtension);
+    }
 
     if (opt.focused)
         image = scrotGrabFocused();
--- a/src/main.c
+++ b/src/main.c
@@ -85,7 +85,8 @@ int main(int argc, char** argv)
         opt.outputFile = strdup("%Y-%m-%d-%H%M%S_$wx$h_scrot.png");
         opt.thumbFile = strdup("%Y-%m-%d-%H%M%S_$wx$h_scrot-thumb.png");
     } else {
-        opt.thumbFile = nameThumbnail(opt.outputFile);
+        if (opt.thumb)
+            opt.thumbFile = optionsNameThumbnail(opt.outputFile);
         scrotHaveFileExtension(opt.outputFile, &haveExtension);
     }
 
--- a/src/options.c
+++ b/src/options.c
@@ -435,13 +435,13 @@ void optionsParseThumbnail(char* optarg)
     }
 }
 
-void optionsParseFileName(char* fileName)
+void optionsParseFileName(char* optarg)
 {
-    opt.outputFile = fileName;
-
-    if (strlen(opt.outputFile) > MAX_OUTPUT_FILENAME)
+    opt.outputFile = strdup(optarg);
+    if (strlen(opt.outputFile) > MAX_OUTPUT_FILENAME) {
         errx(EXIT_FAILURE,"output filename too long, must be "
             "less than %d characters", MAX_OUTPUT_FILENAME);
+    }
 }
 
 void optionsParseNote(char* optarg)
--- a/src/options.c
+++ b/src/options.c
@@ -356,7 +356,7 @@ char* optionsNameThumbnail(const char* name)
     const char* const thumbSuffix = "-thumb";
     const size_t thumbSuffixLength = 7;
     const size_t newNameLength = strlen(name) + thumbSuffixLength;
-    char const* newName = calloc(1, newNameLength);
+    char* newName = calloc(1, newNameLength);
 
     if (!newName)
         err(EXIT_FAILURE, "Unable to allocate thumbnail");
--- a/src/options.c
+++ b/src/options.c
@@ -347,9 +347,6 @@ void optionsParse(int argc, char** argv)
             warnx("unrecognised option %s", argv[optind++]);
     }
 
-    if (opt.thumb && opt.outputFile)
-        opt.thumbFile = nameThumbnail(opt.outputFile);
-
     /* So that we can safely be called again */
     optind = 1;
 }
--- a/src/options.c
+++ b/src/options.c
@@ -351,7 +351,7 @@ void optionsParse(int argc, char** argv)
     optind = 1;
 }
 
-char* nameThumbnail(char* name)
+char* nameThumbnail(const char* name)
 {
     char* extension;
     const char* thumbSuffix = "-thumb";
--- a/src/options.c
+++ b/src/options.c
@@ -442,7 +442,7 @@ void optionsParseThumbnail(char* optarg)
     }
 }
 
-void optionsParseFileName(char* optarg)
+void optionsParseFileName(const char* optarg)
 {
     opt.outputFile = strdup(optarg);
     if (strlen(opt.outputFile) > MAX_OUTPUT_FILENAME) {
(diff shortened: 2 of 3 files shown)


src/main.c
```diff
@@ -84,8 +84,10 @@ int main(int argc, char** argv)
     if (!opt.outputFile) {
         opt.outputFile = strdup("%Y-%m-%d-%H%M%S_$wx$h_scrot.png");
         opt.thumbFile = strdup("%Y-%m-%d-%H%M%S_$wx$h_scrot-thumb.png");
-    } else
+    } else {
+        opt.thumbFile = nameThumbnail(opt.outputFile);
```
daltomi, line 88: Even if the `-t` option is not used, it is always called, it should be:
```C
    } else {
        if (opt.thumb)
            opt.thumbFile = optionsNameThumbnail(opt.outputFile);
        scrotHaveFileExtension(opt.outputFile, &haveExtension);
    }
```

src/options.h
```diff
@@ -72,8 +72,8 @@ struct __ScrotOptions {
 };
 
 void optionsParse(int, char**);
-char* nameThumbnail(char*);
-void optionsParseFileName(char*);
+char* nameThumbnail(const char*);
```
daltomi, line 75: As this function belongs to this file it should continue with the style, something like this:
```C
char* optionsNameThumbnail(const char*);
```

src/options.c
```diff
@@ -442,10 +443,9 @@ void optionsParseThumbnail(char* optarg)
     }
 }
 
-void optionsParseFileName(char* fileName)
+void optionsParseFileName(const char* optarg)
 {
-    opt.outputFile = fileName;
-
+    opt.outputFile = strdup(optarg);
```
daltomi, line 448: Do not reserve memory if it can later fail. Maybe
```C
if (strlen(optarg) > MAX_OUTPUT_FILENAME) {
    ...EXIT..
}
opt.outputFile = strdup(optarg); // or strndup
```

src/options.c
```diff
@@ -347,40 +347,41 @@ void optionsParse(int argc, char** argv)
...
 
-    if (opt.thumb && opt.outputFile)
-        opt.thumbFile = nameThumbnail(opt.outputFile);
-
     /* So that we can safely be called again */
     optind = 1;
 }
 
-char* nameThumbnail(char* name)
+char* nameThumbnail(const char* name)
```
daltomi, line 354: This function can be improved in the following:
- const correctness
- C99: reduce the scope of some variables and use `ptrdiff_t`
- replace malloc with calloc

Proposed patch if you agree:
```C
char* nameThumbnail(const char* name)
{
    const char* const thumbSuffix = "-thumb";
    const size_t thumbPrefixLength = 7;
    const size_t newNameLength = strlen(name) + thumbPrefixLength;
    char *const newName = calloc(1, newNameLength);

    if (!newName)
        err(EXIT_FAILURE, "Unable to allocate thumbnail");

    const char* const extension = strrchr(name, '.');

    if (extension) {
        /* We add one so length includes '\0'*/
        const ptrdiff_t nameLength = (extension - name) + 1;
        strlcpy(newName, name, nameLength);
        strlcat(newName, thumbSuffix, newNameLength);
        strlcat(newName, extension, newNameLength);
    } else
        snprintf(newName, newNameLength, "%s%s", name, thumbSuffix);

    return newName;
}
```

daltomi (changes requested):

src/options.h
daltomi: I understand that you did not write this function, but do me a favor and change the name in your PR, thanks

src/main.c
ifohancroft: Can you please ellaborate?
From what I am understanding, you would like me to put the naming of the thumbFile in an if that's checking if opt.thumb exists? Can there be such a case that it doesn't exist there at the time of calling?

I did this time test the changes with this: ```scrot -t 50 && scrot -t 50 && scrot -t 50 file && scrot -t 50 file && scrot -t 50 file.png && scrot -t 50 file.png && scrot -t 50 file.jpg && scrot -t 50 file.jpg && scrot -t 50 file.jpeg && scrot -t 50 file.jpeg && scrot -t 50 -F file && scrot -t 50 -F file && scrot -t 50 -F file.png && scrot -t 50 -F file.png && scrot -t 50 -F file.jpg && scrot -t 50 -F file.jpg && scrot -t 50 -F file.jpeg && scrot -t 50 -F file.jpeg && scrot -F file -t 50 && scrot -F file -t 50 && scrot -F file.png -t 50 && scrot -F file.png -t 50 && scrot -F file.jpg -t 50 && scrot -F file.jpg -t 50 && scrot -F file.jpeg -t 50 && scrot -F file.jpeg -t 50``` and never ran into a situation that it doesn't exist.

Shouldn't the check if opt.thumb exist be unnecessary?

src/options.h
ifohancroft: Sure!

src/options.c
ifohancroft: I can't agree more. I had it something like this initially, but changed it to the current version to follow what was already done in options.c. It seemed that most functions first assign optarg to their opt.something variable and then check it.

src/options.c
ifohancroft: Thank you! I definitely agree!
The weird thing is I thought I already made the name variable be a constant.

src/main.c
daltomi: I show you first the behavior of the `master` branch:
```gdb
(gdb) b nameThumbnail
Breakpoint 1 at 0x535e: file options.c, line 358.
(gdb) r file.png
[Inferior 1 (process 3032) exited normally]
```
As you can see if the `-t` option is not used the function is never called, which is what is expected.

Now your branch:

```gdb
(gdb) b nameThumbnail
Breakpoint 1 at 0x5361: file options.c, line 355.
(gdb) r file.png

Breakpoint 1, nameThumbnail (name=0x5555555612a0 "file.png") at options.c:355
355     {

```

The function is called even if the `-t` option is not used, and this changes the behavior that it has so far.

ifohancroft: Ah! Thank you! I get what you mean now. You weren't saying that calling the function if there is no -t option will cause the software to throw an error. You don't want the option to be called when there's no need. That makes sense.

ifohancroft:
I am adding the changes now but it's not yet ready as I got some warnings during compilation. It's just so that I don't taint the source, when I make a change, I commit and push it, then in another dir I fetch the changes and compile there.

I'll write here when the changes are done and the warnings are gone.

ifohancroft:
The changes should be done. Let me know if I missed something.

Btw in optionsNameThumbnail, I had to revert newName back to char* newName; otherwise there were warnings during compilation that passing it as the first parameter to strlcpy and strlcat are discarding the 'const' part.

P.S. What is the different between strdup and strndup and in the contexts of scrot, is one preferred over the other?

daltomi (approved):

PR #159, opened 2021-10-20
New feature: blur selection area

daltomi:
New mode for option `--select`: blur
Blurs the selection area.

Ex: `scrot --select=blur`
(_this selection mode is intended to be a complement to the `hide` mode._)

blur


daltomi:
For those who want to test, I am still not sure of the value for the blur.
https://github.com/resurrecting-open-source-projects/scrot/blob/ddab1ce26a2736cf89e4f8ba4f54ec6571051aa3/src/scrot_selection.h#L49
A higher value improves the blurring of the images but the text is destroyed

PR #165, opened 2021-12-02
rename reserved identifiers

guijan:
This patch renames 2 identifiers that use reserved names, and removes a
source file containing typedefs for them.

I see 2 reserved identifiers in use:
```
$ grep -Ri '__' src
src/main.c:#ifdef __i386__
src/note.h:struct __ScrotNote {
src/options.h:struct __ScrotOptions {
src/structs.h:typedef struct __ScrotOptions ScrotOptions;
src/structs.h:typedef struct __ScrotNote ScrotNote;
```
That is, the declarations of the types "struct __ScrotNote" and "struct __ScrotOptions".

To quote a C17 draft (I don't know a web conversion of a C99 draft, but it says the same here) from https://www.iso-9899.info/n1570.html#7.1.3 :
```
-- All identifiers that begin with an underscore and either an uppercase letter or another underscore are always reserved for any use.
```
And further down at https://www.iso-9899.info/n1570.html#7.1.3p2 :
```
If the program declares or defines an identifier in a context in which it is reserved (other than as allowed by 7.1.4), or defines a reserved identifier as a macro name, the behavior is undefined.
```

I'd recommend taking the opportunity to remove [src/structs.h](src/structs.h) entirely (as well as the typedefs it provides), removing the underlines from the structs' names, and sticking to declaring variables of those types with type "struct ScrotOptions" and "struct ScrotNote". That is exactly what this PR does. Actually, I also removed some whitespace, and moved a few includes below the #pragma directive


daltomi:
Remove the reference from here: [Makefile.am#L40](https://github.com/guijan/scrot/blob/f62178dd67ae0ceae5b033ef72bc918ac9077d4f/src/Makefile.am#L40)

PR #170, opened 2021-12-06
bring the logo up to speed and use it

guijan:
Screenshot 2021-12-06 at 07-42-24 guijan scrot at logo-changes
This patch does several things all at once:
- delete scrot-128x128.png

It's not used in the project and we're not a GUI program, so what is it
for?

- move the layer containing the "camera flash" effect in the .xcf logo
to the top layer

It's a bit weird if the camera's flash is behind the camera.

- move the camera flash one pixel to the left

This seems to center it better.

- output the lightly modified logo to doc/scrot.png and add the logo to
the README

I noticed that this project https://github.com/kohler/gifsicle has their
fancy logo at the top of the readme and I thought it was pretty cool. I
think we should do the same.

You can check out how scrot's readme looks with this change at my fork: https://github.com/guijan/scrot/tree/logo-changes


eribertomota:
The logo is used for visual identity. In Debian it is present in DDPO[1] and Salsa[2].

[1] https://qa.debian.org/developer.php?email=eriberto@debian.org
[2] https://salsa.debian.org/debian/scrot

PR #173, opened 2021-12-09
untangle headers

guijan:
2021-12-09_1920x540-TEST

This is a big change and it's doing a lot, I don't expect it to be perfect, but here's a summary and the idea:

- rename src/main.c to src/scrot.c

This isn't merely a pick on the filename. We have a src/scrot.h file (which is also the main offender for the tangled headers) containing the declarations of main.c (among others). To follow the typical C organization where a header's filename only differs from its corresponding .c file in the extension, either scrot.h needed to become main.h or main.c needed to become scrot.c. I follow the convention of naming the source file containing main() after the project, so I renamed main.c to scrot.c.

This actually included touching a few comments which mentioned a "main.c" file.

- untangle src/scrot.h

This file declared pretty much all of the scrot.c functions, which mostly aren't used in other source files, and it also declared functions and globals belonging to other source files (like initXandImlib() and disp from src/imlib.c).
I initially tried removing function declarations one by one and seeing where they were missed and where they were defined until I realized what was going on exactly, and then I simply deleted every single line in the file and rewrote it from scratch. For this reason, I consider that every line in the file is original and now it only contains my copyright.

- untangle all the other headers

With src/scrot.h previously being the central file holding system/Imlib/X11/scrot #includes, the other headers included it to get what they needed, and incidentally they also got a lot of extra baggage.

The most important effect of this is that now the interactions between scrot's source files are strictly defined in code through the headers.
Because the headers only export what other source files call and nothing more, it's easy to know what code can be touched without affecting other source files.
This change has also made the fact that a lot of functions aren't used elsewhere obvious. I noticed it while working on options.c, and I immediately declared showUsage() and showVersion() static and moved them closer to where they are used. I soon realized there was a whole host of other functions that could be made static, so I decided it was out of scope for this patch and left them as they are for now. Every function that is not declared in a header should be made static after this patch, and only functions that are used in other .c files should be declared in a header.

- make every source file include its own header

When your .c file doesn't include its own header, you run into the possibility of a function declaration used by other source files not matching its definition in the .c file. Maintaining this is brittle and requires human oversight which is prone to human error, and failure to do so can invoke undefined behavior.

When your source file includes its own header, the compiler can warn you of mismatching types between declaration and definition.

- remove the AX_PREFIX_CONFIG_H macro from configure.ac

I did it because I considered this part of untangling headers. The macro created a copy of src/config.h with the name src/scrot_config.h.
Here's the autotools documentation of the macro: https://www.gnu.org/software/autoconf-archive/ax_prefix_config_h.html

To quote the documentation, the macro is used to "Generate an installable config.h." And "A package should not normally install its config.h as a system header, but if it must, this macro can be used to avoid namespace pollution by making a copy of config.h with a prefix added to all the macro names."
scrot isn't a library and we don't install our header, so I removed it.

Code using the C macros provided by autotools should #include "config.h" now, and the parts of the code that do have been updated to match this.

- create a src/imlib.h header corresponding to src/imlib.c

All of imlib.c's declarations were in scrot.h, so it didn't have its own header. Now that scrot.h doesn't do the business of other source files, I created a src/imlib.h. The comments that were near the declarations of imlib.c's globals were moved to imlib.c.

- make every .h file #include what it needs

When I proposed following the WebKit style in scrot, I tried automatically converting the source, but then I noticed that this introduced a compilation failure (and only after I sent the patch, whoops). clang-format sorted the #includes alphabetically, but the headers depended in their order of inclusion, and this shouldn't happen.
It happened because not all headers #included what they needed, and they happened to get it from other headers when a .c file #included them.

Now all the headers #include only what they need to get the types in their declarations, they have only what they need and nothing more, so scrot headers can't depend on a specific order of inclusion anymore.

- make every .c file #include what it needs

Again, previously, everything was #included through scrot.h. An obvious bug that could arise from this is deleting an #include from scrot.h and later finding out that an unrelated .c file breaks. This shouldn't happen.

The inclusion order I used is "<sys/*> headers -> C/POSIX/BSD headers -> libraries -> local headers" with each group of headers alphabetically sorted and separated by a blank line.

- misc small changes

I deleted some whitespace and the "This file is part of the scrot project" comments. I also organized the list of sourced files in src/Makefile.am a bit better.

I've build tested this change on OpenBSD and I've tested the resulting binary after the changes with this command line:
```
./scrot '%Y-%m-%d_$wx$h-TEST.png' -e 'cp $f out.png' -s -d 2 -f -c -b -l style=dash,width=3,color="red" -p -o -n "-f '/usr/X11R6/lib/X11/fonts/TTF/DejaVuSans.ttf/40' -x 10 -y 20 -c 255,0,0,255 -t 'Hi'"
```
I got the command line from:
https://marc.info/?l=openbsd-ports&m=161139638513700&w=2
I made a small change to it, the resulting png file is not moved but copied, so you can check if the formatted filename looks right.
As you can see, it tries a lot of options at the same time, to exercise the most code possible. We don't have automated testing so it should be done by hand.
I'm annexing the screenshot taken by that command in the PR.

You can use that command to test the patch on your system by changing "/usr/X11R6/lib/X11/fonts/TTF/DejaVuSans.ttf" to the path of any font if you don't have a font in that exact path.

### DIFF
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -15,7 +15,7 @@ Copyright 2020      nothub
 Copyright 2020      Sean Brennan <zettix1@gmail.com>
 Copyright 2021      c0dev0id <sh+github@codevoid.de>
 Copyright 2021      Christopher R. Nelson <christopher.nelson@languidnights.com>
-Copyright 2021      Guilherme Janczak <guilherme.janczak@yandex.com>
+Copyright 2021-2022 Guilherme Janczak <guilherme.janczak@yandex.com>
 Copyright 2021      IFo Hancroft <contact@ifohancroft.com>
 Copyright 2021      Peter Wu <peterwu@hotmail.com>
 
@@ -65,39 +65,24 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #include "slist.h"
 #include "util.h"
 
-size_t scrotHaveFileExtension(char const *, char **);
-Imlib_Image scrotGrabFocused(void);
-Imlib_Image scrotGrabAutoselect(void);
-Imlib_Image scrotGrabShotMulti(void);
-Imlib_Image scrotGrabStackWindows(void);
-Imlib_Image scrotGrabShot(void);
-void scrotCheckIfOverwriteFile(char **);
-void scrotExecApp(Imlib_Image, struct tm *, char *, char *);
-char *imPrintf(char *, struct tm *, char *, char *, Imlib_Image);
-Window scrotGetClientWindow(Display *, Window);
-Window scrotFindWindowByProperty(Display *, const Window, const Atom);
-Imlib_Image stalkImageConcat(ScrotList *, enum Direction const);
+static void uninitXAndImlib(void);
+static size_t scrotHaveFileExtension(char const *, char **);
+static Imlib_Image scrotGrabFocused(void);
+static void applyFilterIfRequired(void);
+static Bool scrotXEventVisibility(Display *, XEvent *, XPointer);
+static Imlib_Image scrotGrabAutoselect(void);
+static Imlib_Image scrotGrabShotMulti(void);
+static Imlib_Image scrotGrabStackWindows(void);
+static Imlib_Image scrotGrabShot(void);
+static void scrotCheckIfOverwriteFile(char **);
+static void scrotExecApp(Imlib_Image, struct tm *, char *, char *);
+static char *imPrintf(char *, struct tm *, char *, char *, Imlib_Image);
+static Window scrotGetClientWindow(Display *, Window);
+static Window scrotFindWindowByProperty(Display *, const Window,
+                                              const Atom);
+static Imlib_Image  stalkImageConcat(ScrotList *, enum Direction const);
 
-/* atexit register func. */
-static void uninitXAndImlib(void)
-{
-    if (opt.note)
-        scrotNoteFree();
-
-    if (disp) {
-        XCloseDisplay(disp);
-        disp = NULL;
-    }
-}
-
-// It assumes that the local variable 'scrot.c:Imlib_Image image' is in context
-static void applyFilterIfRequired(void)
-{
-    if (opt.script)
-        imlib_apply_filter(opt.script);
-}
-
-int main(int argc, char** argv)
+int main(int argc, char *argv[])
 {
     Imlib_Image image;
     Imlib_Image thumbnail;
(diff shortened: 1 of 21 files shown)


guijan:
I added a 2nd commit to the branch of this PR, it removes the autoconf-archive dependency because the 1st commit removes the need for it.

daltomi:
Hi @guijan 
In general I agree with the changes, I will see it better after the holidays.

A note about your authorship:
The types of variables and their names, the names of functions as well as the types of parameters and number of parameters, have authorship over time.
You only have to indicate in a comment that part of the code comes from X file or files.

See https://github.com/resurrecting-open-source-projects/scrot/blob/d0f5bf4b99583e2a9a79b6c474d26dd17c4942c7/src/scrot_selection.c#L31

 If you do not indicate that, it is explicitly understood that you and only you are the author of the entire file, which is not true.

guijan:
I added the notice and squashed the commits.
I also changed more references to a "main.c" file to "scrot.c". Seems I didn't spot all the comments previously.

src/scrot.c
```diff
@@ -40,8 +40,43 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
...
+Imlib_Image scrotGrabAutoselect(void);
+Imlib_Image scrotGrabShotMulti(void);
+Imlib_Image scrotGrabStackWindows(void);
+Imlib_Image scrotGrabShot(void);
+void scrotCheckIfOverwriteFile(char **);
+void scrotExecApp(Imlib_Image, struct tm *, char *, char *);
+char *imPrintf(char *, struct tm *, char *, char *, Imlib_Image);
+Window scrotGetClientWindow(Display *, Window);
+Window scrotFindWindowByProperty(Display *, const Window, const Atom);
+Imlib_Image stalkImageConcat(ScrotList *, enum Direction const);
```
daltomi, line 79: These functions must be specified as `static` (declaration and definition).

guijan: I made those static, and I also went through the other files and found a few more that could be made static. There were also some style fixes. Check the text of the 2nd commit of the latest force-push (I had to due to a conflict with master)

src/selection_edge.c
```diff
@@ -146,3 +125,44 @@ void selectionEdgeMotionDraw(int x0, int y0, int x1, int y1)
...
+            break;
+        nanosleep(&delay, NULL);
+    }
+}
+
+
+static Bool xeventUnmap(Display *dpy, XEvent *ev, XPointer arg)
+{
+    (void)dpy; // unused
+    Window *win = arg;
```
daltomi, line 166: Generic pointers must be cast so the intention is well documented. I don't know why you changed it.
I'm going to change it before the merge.

daltomi (approved):

PR #175, opened 2021-12-20
build: respect docdir

thesamesam:
Before this patch with --docdir=/usr/share/doc/scrot-1.7-r1 (example downstream
versioning for patches):
```
>>> /usr/share/doc/scrot-1.7-r1/
>>> /usr/share/doc/scrot-1.7-r1/README.md.bz2
>>> /usr/share/doc/scrot-1.7-r1/ChangeLog.bz2
>>> /usr/share/doc/scrot-1.7-r1/AUTHORS.bz2
>>> /usr/share/doc/scrot/
>>> /usr/share/doc/scrot/scrot.png
>>> /usr/share/doc/scrot/README.md.bz2
>>> /usr/share/doc/scrot/ChangeLog.bz2
>>> /usr/share/doc/scrot/AUTHORS.bz2
```

After:
```
>>> /usr/share/doc/scrot-1.7-r1/scrot.png
>>> /usr/share/doc/scrot-1.7-r1/README.md.bz2
>>> /usr/share/doc/scrot-1.7-r1/ChangeLog.bz2
>>> /usr/share/doc/scrot-1.7-r1/AUTHORS.bz2
```

This avoids installing duplicate files & installs them to the
same location but it's overridable by --docdir in ./configure which
helps downstreams.

Bug: https://bugs.gentoo.org/828928
Signed-off-by: Sam James <sam@gentoo.org>

PR #177, opened 2022-01-09
small CI revamp

guijan:
- Add Cygwin to the CI

This is mainly to test portability further.
Cygwin uses the newlib libc on top of Windows. The newlib libc implements a lot of BSD functions. It's a weird platform so maybe it'll catch a bug in the future.

- Add distcheck CI target

I noticed that in PR #175 @daltomi noticed the PR broke `make distcheck`, this could have been caught by the CI, now it will. I also noticed that the Github Actions CI was doing almost all of https://www.gnu.org/software/automake/manual/html_node/Checking-the-Distribution.html by hand, so the lines were deleted as `make distcheck` now fills their role.

- Bring the Cirrus CI in line with Github Actions

The Cirrus actions I wrote didn't run `scrot -v` or do the equivalent of `make distcheck`, now they do. I had to add the gzip package to the Alpine Linux VM as distcheck depends on it.

- Update MacOS version in Cirrus CI

- Update actions/checkout in Github Actions CI

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -24,7 +24,10 @@ common_template: &common_script
   build_script:
     - ./autogen.sh
     - ./configure
+    - make distcheck
+  test_script:
     - make
+    - src/scrot -v
   on_failure:
     autotools_artifacts:
      path: config.log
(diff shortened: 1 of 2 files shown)


daltomi:
> This is mainly to test portability further.
> Cygwin uses the newlib libc on top of Windows. The newlib libc implements a lot of BSD functions. It's a weird platform so maybe it'll catch a bug in the future.

I do not understand this.
We already have FreeBSD in the CI, why do we need that strange platform to detect BSD errors?

guijan:
It's not necessarily to catch errors around BSD functions in specific, it's an unusual platform. 

Sometimes being portable to wildly different platforms can expose bugs. Different implementations of the same API are going to have slightly different semantics within the constraints of the API's definition. They might also break our assumptions that are outside of any standard.

PR #180, opened 2022-01-11
add link time optimization and warnings by default

guijan:
Quoting
[GCC Documentation](https://gcc.gnu.org/wiki/LinkTimeOptimization):
> Link Time Optimization (LTO) gives GCC the capability of dumping its
> internal representation (GIMPLE) to disk, so that all the different
> compilation units that make up a single executable can be optimized as
> a single module. This expands the scope of inter-procedural
> optimizations to encompass the whole program (or, rather, everything
> that is visible at link time).

It also has the nice side effect of allowing warnings across translation
units, as attested by these internet blogs:
[gcc 8: link time and interprocedura optimization](https://hubicka.blogspot.com/2018/06/gcc-8-link-time-and-interprocedural.html)
[gnu link time optimization finds non-matching declarations](https://mcuoneclipse.com/2018/05/31/gnu-link-time-optimization-finds-non-matching-declarations/)
> A new user-visible feature is warning about type mismatches across
> units.

Some Linux distros are migrating towards enabling it by default for all
packages, as [Phoronix](https://www.phoronix.com/scan.php?page=news_item&px=Ubuntu-21.04-LTO-Packages) claims. Some OSes (like OpenBSD) don't touch the compile flags of their packages though, so we should do it ourselves.

This change adds autoconf-archive back as a dependency because I used autoconf-archive macros to get it done. Sorry, I only had the idea after the commit that removed it was pushed to master.

This change replaces the CFLAGS with "-O2 -flto -g" and appends the same flags to LDFLAGS (LTO requires this) if the user hasn't set CFLAGS, effectively changing AC_PROG_CC's default of "-O2 -g". I checked out scrot 1.6, 1.7, and commit 6d299f5 (current master as of writing this), and compiled all 3 of them with CFLAGS='-O2', copying each resulting binary to /tmp/. Then I compiled 6d299f5 with CFLAGS='-O2 -flto' to show the difference made by this change:

```console
$ size /tmp/scrot-*
text    data    bss     dec     hex
45417   2952    392     48761   be79    /tmp/scrot-1.6
43724   3232    208     47164   b83c    /tmp/scrot-1.7
42592   3168    208     45968   b390    /tmp/scrot-6d299f5
39649   2880    200     42729   a6e9    /tmp/scrot-6d299f5-flto
```

There's an overall trend with scrot getting 5% smaller with all the cleanups that have been made so far, and now LTO on top of 6d299f5 reduces the binary by a further 7% for a total 12% reduction in binary size. LTO and code quality aren't purely there for smaller binaries, but smaller executables are an easy way to notice them.

I also turned on all warnings by default.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -49,9 +49,10 @@ task:
       - CC: clang
       - CC: gcc
   install_script:
-    - apk add autoconf automake gzip make pkgconfig clang gcc xorg-server-dev
-              libxcomposite-dev libxext-dev libxfixes-dev imlib2-dev musl-dev
-              bsd-compat-headers
+    - apk add autoconf autoconf-archive automake binutils gzip make pkgconfig
+                       gcc clang xorg-server-dev libxcomposite-dev libxext-dev
+                       libxfixes-dev imlib2-dev musl-dev bsd-compat-headers
+                       libbsd-dev llvm-dev
   << : *common_script
 
 task:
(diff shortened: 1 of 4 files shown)


daltomi:
I like it, I've been using it for years.

PR #181, opened 2022-01-13
Add new format specifier: $W

daltomi:
The FILE parameters can take a new format specifier: $W
The name of the window (only for --select and --focused)

Ex:
    $ scrot --focused mytitle$W.png


guijan (commented):
We have a lot of global state and adding a feature means being aware of the entire source code and making changes in multiple different places, I recommend that adding a feature should come with adapting scrot's design to it

PR #192, opened 2022-06-11
remove i386 ifdef

guijan:
I finally understood what that #ifdef was: it was a micro optimization.
The #ifdef avoids a copy if xcim->pixels happens to be uint32, this is
the case on 32-bit Unix. That's the last #ifdef in scrot being removed.
I want to do this because it's easier to make sure scrot works on
various platforms if they're all running the same code.

This also adds a dependency on my libobsd library under OS X because
scrot now uses reallocarray() which isn't provided by OS X. My library
isn't packaged by Brew.

### DIFF
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -413,7 +413,7 @@ void scrotGrabMousePointer(const Imlib_Image image, const int xOffset,
     DATA32 *pixels;
     size_t i;
 
-    if ((pixels = reallocarray(NULL, width * height, sizeof(pixels))) == NULL)
+    if ((pixels = reallocarray(NULL, width * height, sizeof(*pixels))) == NULL)
         err(EXIT_FAILURE, "reallocarray");
     /*
      * Copy (unsigned long *)xcim->pixels into (DATA32 *)pixels by assignment
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -401,7 +401,6 @@ void scrotGrabMousePointer(const Imlib_Image image, const int xOffset,
     const int yOffset)
 {
     XFixesCursorImage *xcim = XFixesGetCursorImage(disp);
-
     if (!xcim) {
         warnx("Failed to get mouse cursor image.");
         return;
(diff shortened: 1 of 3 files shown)


src/scrot.c
```diff
@@ -411,24 +410,20 @@ void scrotGrabMousePointer(const Imlib_Image image, const int xOffset,
...
-#ifdef __i386__
-    pixels = (DATA32 *)xcim->pixels;
-#else
-    DATA32 data[width * height * 4];
-
-    unsigned int i;
+    DATA32 *pixels;
+    size_t i;
+
+    if ((pixels = reallocarray(NULL, width * height, sizeof(pixels))) == NULL)
```
daltomi, line 416: Must be `sizeof(*pixels)`

src/scrot.c
```diff
@@ -411,24 +410,20 @@ void scrotGrabMousePointer(const Imlib_Image image, const int xOffset,
...
+        err(EXIT_FAILURE, "reallocarray");
```
daltomi, line 417: Before exiting you must call `XFree(xcim)` because it is memory allocated by a library and we have no control over it.

src/scrot.c
guijan: Good eye.

src/scrot.c
guijan: The process' memory is deallocated when it exits.

guijan:
Please confirm whether leaving memory deallocation to proccess termination in the error case is acceptable for scrot or not, I want to get rid of that last #ifdef.

daltomi:
> I want to do this because it's easier to make sure scrot works on
> various platforms if they're all running the same code.

You say that and then use a custom library to make this commit work on a platform that currently works correctly.
I cannot allow the use of collaborators' personal libraries.

I have already [answered you here,](https://github.com/guijan/scrot/commit/c9ccb79215d4c60f9b291ee8391be45748581bf7) what I think of this commit, and I never received an answer from you.


> Please confirm whether leaving memory deallocation to proccess termination in the error case is acceptable for scrot or not, I want to get rid of that last #ifdef.

With the Xlib and Imlib2 libraries the policy is to release explicitly. True, this is not always the case, but this `scrotGrabMousePointer` function before this commit did not have a memory leak in Xlib.

PR #197, opened 2022-09-06
fix shift error

N-R-K:
the PIXEL_ARGB() macro would do an overwide shift if `int` is 16bits.
and in case of 32bit ints the result of the shift `a << 24` can
overflow, which is undefined behavior.

test case, (compile with UBSan):

	$ ./configure CFLAGS="-fsanitize=undefined -g"
	$ make
	$ ./src/scrot -shide,some.png -l "opacity=254" -o /tmp/whatevr.png
	scrot_selection.c:426:25: runtime error: left shift of 156 by 24 places cannot be represented in type 'int'

this patch also removes the PIXEL_* macros entirely, they weren't needed
because only the alpha changed, r,g,b was unchanged.

### DIFF
--- a/src/scrot_selection.c
+++ b/src/scrot_selection.c
@@ -5,6 +5,7 @@ Copyright 2021       Martin C <martincation@protonmail.com>
 Copyright 2021       Peter Wu <peterwu@hotmail.com>
 Copyright 2021       Wilson Smith <01wsmith+gh@gmail.com>
 Copyright 2021-2022  Guilherme Janczak <guilherme.janczak@yandex.com>
+Copyright 2022       NRK <nrk@disroot.org>
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


daltomi:
Very good thanks!

I need you to include your `Copyright` in this PR.

N-R-K:
> Very good thanks!
> 
> I need you to include your `Copyright` in this PR.

Done https://github.com/resurrecting-open-source-projects/scrot/pull/197/commits/09772ef761957ed6d0578416599a702fcef1f424

---

Also one more thing, as pointed out in [here](https://github.com/resurrecting-open-source-projects/scrot/pull/193#issuecomment-1237529245), the `DATA` types are depricated by imlib2. Do you prefer I change `DATA32` to `uint32_t` in this PR or open a new PR for that change?

daltomi:
I see. Yes, do it here.

PR #198, opened 2022-09-06
remove VLA and imlib2's deprecated DATA32 type

guijan:
@N-R-K pointed out the deprecated type and revived interest in removing
the only usage of a VLA in this project.

See the discussion after #193 was merged.
I've tested taking a screenshot with and without a mouse cursor after applying this change, it worked fine.

### DIFF
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -396,29 +396,23 @@ Window scrotGetWindow(Display *display, Window window, int x, int y)
     return target;
 }
 
-
 void scrotGrabMousePointer(const Imlib_Image image, const int xOffset,
     const int yOffset)
 {
     XFixesCursorImage *xcim = NULL;
-    unsigned short width;
-    unsigned short height;
-    int x;
-    int y;
-    size_t pixcnt;
     uint32_t *pixels = NULL;
     Imlib_Image imcursor;
-
     /* Get the X cursor and the information we want. */
     if ((xcim = XFixesGetCursorImage(disp)) == NULL) {
         warnx("Can't get the cursor from X");
         goto end;
     }
-    width = xcim->width;
-    height = xcim->height;
-    x = (xcim->x - xcim->xhot) - xOffset;
-    y = (xcim->y - xcim->yhot) - yOffset;
-    pixcnt = (size_t)width*height;
+    const unsigned short width = xcim->width;
+    const unsigned short height = xcim->height;
+    const int x = (xcim->x - xcim->xhot) - xOffset;
+    const int y = (xcim->y - xcim->yhot) - yOffset;
+    const size_t pixcnt = (size_t)width*height;
+
     /* xcim->pixels wouldn't be valid if sizeof(*pixels)*pixcnt wrapped around.
      */
     if ((pixels = malloc(sizeof(*pixels)*pixcnt)) == NULL) {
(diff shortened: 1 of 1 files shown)


src/scrot.c
```diff
@@ -400,38 +400,54 @@ Window scrotGetWindow(Display *display, Window window, int x, int y)
...
-
-    Imlib_Image imcursor = imlib_create_image_using_data(width, height, data);
-
-    XFree(xcim);
-
+    width = xcim->width;
+    height = xcim->height;
+    x = (xcim->x - xcim->xhot) - xOffset;
+    y = (xcim->y - xcim->yhot) - yOffset;
+    pixcnt = width*height;
```
N-R-K, line 422: Since `width` and `height` are `unsigned short`, the multiplication takes place in `int` after integer promotion. Cast one of the operands to `size_t` so that the mul happens at `size_t`.

```suggestion
    pixcnt = (size_t)width*height;
```

src/scrot.c
```diff
@@ -400,38 +400,54 @@ Window scrotGetWindow(Display *display, Window window, int x, int y)
...
-        return;
+    XFixesCursorImage *xcim = NULL;
+    char *errmsg = NULL;
+    unsigned short width;
+    unsigned short height;
+    int x;
+    int y;
+    size_t pixcnt;
+    uint32_t *pixels = NULL;
+    Imlib_Image imcursor;
```
N-R-K, line 411: Aside from the pointers which gets freed at the cleanup block, why not declare rest of the variables where they are actually initialized.

N-R-K (changes requested):

src/scrot.c
```diff
@@ -400,38 +400,54 @@ Window scrotGetWindow(Display *display, Window window, int x, int y)
...
+
+    /* Convert an X cursor into the format imlib wants. */
+    if ((xcim = XFixesGetCursorImage(disp)) == NULL) {
+        errmsg = "Can't get the cursor from X";
```
N-R-K, line 415: Also this entire `errmsg` seems completely unnecessary. Why not just do the following:

```c
if (...) {
        warn("msg");
        goto end;
}
```

src/scrot.c
```diff
@@ -400,38 +400,54 @@ Window scrotGetWindow(Display *display, Window window, int x, int y)
...
+    char *errmsg = NULL;
+    unsigned short width;
+    unsigned short height;
+    int x;
+    int y;
+    size_t pixcnt;
+    uint32_t *pixels = NULL;
+    Imlib_Image imcursor;
+
+    /* Convert an X cursor into the format imlib wants. */
```
N-R-K, line 413: Misleading comment. This comment should appear before the loop at line 429.

src/scrot.c
guijan: Sure.

src/scrot.c
guijan: No reason, just a style choice.

guijan: Hi again, is keeping the original location of the declarations important or is it okay if this PR changes them?

N-R-K: No clue, I'm not the maintainer here :shrug:

daltomi:
>
> Hi again, is keeping the original location of the declarations important or is it okay if this PR changes them? 
> 
Hi,

Try reducing the scope of variables and using 'const-correctness' if possible.

daltomi:
I made a few more small changes, I understand you wouldn't have a problem with this. I merge the branch in a few moments.

PR #199, opened 2022-10-18
Add -M/--monitor option to capture a Xinerama display

zevweiss:
This is based on the patches in PR #137 by Daniel T. Borelli (daltomi), rebased, tidied up a bit and with a man page update.

Fixes: #38


src/scrot.c
```diff
@@ -860,6 +864,38 @@ static Imlib_Image scrotGrabShotMulti(void)
...
 
+static Imlib_Image scrotGrabShotMonitor(void)
+{
+    int eventBase, errBase;
+
+    if (!XineramaQueryExtension(disp, &eventBase, &errBase))
+        err(EXIT_FAILURE, "Xinerama extension not found");
+
+    int numScreens = 0;
+    XineramaScreenInfo* screens = XineramaQueryScreens(disp, &numScreens);
```
daltomi, line 875: TODO:

- [ ] Some of the code is pre-style WebKit. [Put the * next to the variable name and not the type](https://webkit.org/code-style-guidelines/#pointers-non-cpp)(`screens` and `mon`)
- [ ] Add your `Copyright` in the .h and .c files that you have changed.

daltomi (changes requested):

daltomi:
Hi @zevweiss 
It looks good.
Before merging with `master` I would like more contributors to test this PR; one that has 2 or more monitors.

TODO for myself:

- [ ] [README.md: Indicate the reference to the new library.]( https://github.com/resurrecting-open-source-projects/scrot#building
) Ref: [libXinerama](https://gitlab.freedesktop.org/xorg/lib/libxinerama)

src/scrot.c
zevweiss: Sure, now done.

daltomi (approved):

N-R-K:
Since multi-monitor isn't a common case, I think it makes sense to have an option to compile this feature out for those who might not want the extra dependency.

Looking at the code, I don't think it'd require too much ugly `ifdefs` to achieve. Just guarding the include and the function body should do:

```c
#ifdef ENABLE_MULTIMONITOR
    #include <X11/extensions/Xinerama.h>
#endif

/* ... */

static Imlib_Image scrotGrabShotMonitor(void)
{
#ifdef ENABLE_MULTIMONITOR
    /* ... */
#else
    errx(EXIT_FAILURE, "scrot was compiled without mutli-monitor support);
#endif 
}
```

zevweiss:
> Since multi-monitor isn't a common case, I think it makes sense to have an option to compile this feature out for those who might not want the extra dependency.

I dunno -- multi-head setups are probably a minority, sure, but I don't think they're a particularly tiny one.  Laptop users frequently plug in to an external display in a home/office desk environment or when presenting something via a projector, for example.

Also, while libxinerama is strictly speaking an extra dependency, it's already a dependency of gtk and most desktop environments and window managers (even pretty bare-bones ones like fluxbox/openbox/fvwm, say), so I'd guess the number of systems that have scrot's other dependencies but _don't_ already have it is probably pretty vanishingly small.

src/options.c
```diff
@@ -488,8 +494,8 @@ static void showUsage(void)
 {
     fputs(/* Check that everything lines up after any changes. */
         "usage:  " PACKAGE " [-bcfhimopuvz] [-a X,Y,W,H] [-C NAME] [-D DISPLAY]\n"
-        "              [-d SEC] [-e CMD] [-F FILE] [-k OPT] [-l STYLE] [-n OPTS]\n"
-        "              [-q NUM] [-S CMD] [-s OPTS] [-t NUM | GEOM] [FILE]\n",
+        "              [-d SEC] [-e CMD] [-F FILE] [-k OPT] [-l STYLE] [-M MONITOR]\n"
```
daltomi, line 497: I prefer to stick with the other options: use `NUM` for an argument that expects a number.
With the word `MONITOR` someone could get confused if they have to enter a string  like: `VGA-1`, or `LVDS1` or similar.

zevweiss: Sure; done.

PR #204, opened 2023-01-05
Automake: Use default compression format

daltomi:
Remove the forced bzip2 format and use only the default: gzip

Doc: make dist

Generate a ‘gzip’ tar archive of the distribution.
This is the only format enabled by default. 
By default, this rule makes gzip use a compression option of --best.
To make it use a different one, set the GZIP_ENV environment variable.
For example, ‘make dist-gzip GZIP_ENV=-7’.


daltomi:
It also fixes an error in the build of Ciruss CI: `debian-*`

```bash
/bin/bash: line 1: bzip2: command not found
make[1]: *** [Makefile:650: dist-bzip2] Error 127
```

PR #206, opened 2023-01-06
add informative error messages when parsing args

guijan:
This change also comes with some under the hood modifications to how args are handled.

This is a continuation of #200.

I've only gone as far as getting a build success, I haven't actually tested the changes. 
I'm leaving it a draft, I'm going to bed soon.

### DIFF
--- a/src/note.c
+++ b/src/note.c
@@ -1,7 +1,7 @@
 /* note.c
 
 Copyright 2019-2022 Daniel T. Borelli <danieltborelli@gmail.com>
-Copyright 2021-2022 Guilherme Janczak <guilherme.janczak@yandex.com>
+Copyright 2021-2023 Guilherme Janczak <guilherme.janczak@yandex.com>
 Copyright 2021      IFo Hancroft <contact@ifohancroft.com>
 Copyright 2021      Peter Wu <peterwu@hotmail.com>
 
@@ -28,6 +28,7 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 #include <assert.h>
 #include <err.h>
+#include <limits.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -114,6 +115,7 @@ void scrotNoteNew(char const *const format)
         const char type = *++token;
         char *savePtr = NULL;
         char *c;
+        const char *errmsg;
 
         nextSpace(&token);
 
@@ -129,7 +131,12 @@ void scrotNoteNew(char const *const format)
             if (!number)
                 errx(EXIT_FAILURE, "Error --note option : Malformed syntax for -f, required number.");
 
-            const int fontSize = optionsParseRequiredNumber(++number);
+            const int fontSize = optionsParseNum(++number, 1, INT_MAX,
+                &errmsg);
+            if (errmsg) {
+                    errx(EXIT_FAILURE, "option --note: font size '%s' is %s",
+                        number, errmsg);
+            }
 
             if (fontSize < 6)
                 warnx("Warning: --note option: font size < 6");
@@ -161,9 +168,17 @@ void scrotNoteNew(char const *const format)
 
             while (c) {
                 token = c;
+                char *const space = strchr(c, ' ');
 
-                const int color = optionsParseRequireRange(
-                        optionsParseRequiredNumber(c), 0 ,255);
+                if (space)
+                    *space = '\0';
+                const int color = optionsParseNum(c, 0, 255, &errmsg);
+                if (errmsg) {
+                        errx(EXIT_FAILURE, "option --note: color '%s' is %s", c,
+                            errmsg);
+                }
+                if (space)
+                    *space = ' ';
 
                 switch (++numberColors) {
                 case 1:
--- a/src/options.c
+++ b/src/options.c
@@ -13,7 +13,7 @@ Copyright 2019-2022 Daniel T. Borelli <danieltborelli@gmail.com>
 Copyright 2019      Jade Auer <jade@trashwitch.dev>
 Copyright 2020      Sean Brennan <zettix1@gmail.com>
 Copyright 2021      Christopher R. Nelson <christopher.nelson@languidnights.com>
-Copyright 2021-2022 Guilherme Janczak <guilherme.janczak@yandex.com>
+Copyright 2021-2023 Guilherme Janczak <guilherme.janczak@yandex.com>
 Copyright 2021      IFo Hancroft <contact@ifohancroft.com>
 Copyright 2021      Peter Wu <peterwu@hotmail.com>
 Copyright 2021      Wilson Smith <01wsmith+gh@gmail.com>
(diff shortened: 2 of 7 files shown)


src/options.c
```diff
@@ -255,16 +261,12 @@ static void optionsParseLine(char *optarg)
...
-            }
-
-            opt.lineWidth = optionsParseRequiredNumber(value);
-
-            if (opt.lineWidth <= 0 || opt.lineWidth > 8) {
-                errx(EXIT_FAILURE, "Value of the range (1..8) for "
-                    "suboption '%s': %d", token[Width], opt.lineWidth);
+            opt.lineWidth = options_parsenum(value, 1, 8, &errmsg);
+            if (errmsg) {
+                if (value == NULL)
```
daltomi, line 266: When `--line width` then `value == NULL`

When `--line width=` then `value[0] = '\0'`

Two possible solutions:
 ```c
if (value == NULL || value[0] == '\0')
```

```c
if (!optionsParseIsString(value))
```

daltomi:
To fix:  `scrot --line width=a` is preferable that it indicates that `a` it is not a number.
Master branch:
```bash 
scrot: the option is not a number: a
```
This branch:
```bash 
scrot: option --line: suboption 'width': 'a' is too small
```

daltomi:
Code style, name of functions: https://webkit.org/code-style-guidelines/#names
>Lower-case the first letter, including all letters in an acronym, in a variable or function name.

src/options.c
```diff
@@ -293,11 +295,14 @@ static void optionsParseLine(char *optarg)
             opt.lineMode = estrdup(value);
             break;
         case Opacity:
-            if (!optionsParseIsString(value)) {
-                errx(EXIT_FAILURE, "Missing value for suboption '%s'",
-                    token[Opacity]);
+            opt.lineOpacity = options_parsenum(value,
+                SELECTION_EDGE_OPACITY_MIN, SELECTION_OPACITY_MAX, &errmsg);
```
daltomi, line 299: There are 2 separate opacity ranges, one for the `scrot_selection.c` file and the other for the `selection_edge.c` file.
They must be independent.

src/options.c
```diff
@@ -563,30 +585,31 @@ void optionsParseDisplay(char *optarg)
 void optionsParseThumbnail(char *optarg)
```
daltomi, line 585: It's broken, this should work like `master`: `scrot -t 20`
Wrong result: `scrot: option --thumb: resolution height is missing`

src/options.c
guijan: Hmm, I copypasted that from the original. `grep` says that in the current master, SELECTION_OPACITY_MIN and SELECTION_EDGE_OPACITY_MAX are unused. 
This seems to be because opt.lineOpacity is reused across "classic" and "edge" selection modes.

Maybe it should be left as is and fixed in a separate PR?

src/options.c
guijan: `options_parse()` handles that
2023-01-06-145454_624x164_scrot

guijan:
Going to take the chance to force-push while nobody is looking.
So the current situation is that this draft needs a reply and more testing from @daltomi who is a good tester, but it's pretty much complete.

src/options.c
daltomi: I only see that `SELECTION_EDGE_OPACITY_MAX` is not used.

#### Old behavior
I will try to explain the current behavior: the highs and lows for `edge` mode is so that the selection area is **not invisible**, as it is not in `classic` mode either.
These ranges are internal to scrot and are not the responsibility of the user. The only thing the user has to know, by reading the manpage, is that the opacity range is between 0 and 255. That's  why functions like `optionsParseRequireRange` and `nonNegativeNumber` never signal an error to the user, because it's for use internal scrot.

#### New behavior

Ok, let's change the behavior of the `edge` selection mode, you have to change `SELECTION_EDGE_OPACITY_MIN` to `SELECTION_OPACITY_MIN` and remove all `_EDGE_OPACITY` references.
```suggestion
                SELECTION_OPACITY_MIN, SELECTION_OPACITY_MAX, &errmsg);
```

 This changes the old behavior of the `edge` selection mode: if the opacity is equal to 0 it draws the selection area with opacity 10 which makes it **always visible**, as it is with the `classic` selection mode, but with this change if the user uses opacity 0 the `edge` selection mode will be **invisible** (with CWM enabled of course).

src/options.c
daltomi: Now there is another case, it does not generate the thumbnail image file with: `scrot -t 100x100`

daltomi:
```diff
             errx(EXIT_FAILURE, "option --thumb: resolution height '%s' is %s",
                 height, errmsg);
         }
+
+        // allow create thumbnail
+        opt.thumb = 1;
+
     } else {
         /* optarg is a percentage. */
         opt.thumb = options_parsenum(optarg, 1, INT_MAX, &errmsg);
```

src/options.c
```diff
@@ -533,24 +561,28 @@ char *optionsNameThumbnail(const char *name)
...
     int i = 0;
+    const char *errmsg;
+
+    /* Geometry dimensions must be in format x,y,w,h */
+    token = strtok(optarg, ",");
+    for (; token != NULL; token = strtok(NULL, ",")) {
+        if (dimensions[i] == NULL)
+            errx(EXIT_FAILURE, "option --autoselect: too many dimensions");
+
+        *dimensions[i] = options_parsenum(token, 1, INT_MAX, &errmsg);
```
daltomi, line 575: On the `master` branch you can use `autoselect` from range 0: `scrot -a 0,0,10,10`
The minor range must start at 0.

Making the change in both branches gives the same result.
```bash
382ee633c173f1ed1f44f61b7cf54a8e  2023-01-06-182253_10x10_scrot.png
382ee633c173f1ed1f44f61b7cf54a8e  2023-01-06-182319_10x10_scrot.png <-- master

**Comment by daltomi** [comment]
Update Copyright: `scrot.c` and `scrot_selection.c`

File: src/note.c [hunk 5]
```diff
@@ -162,8 +169,11 @@ void scrotNoteNew(char const *const format)
             while (c) {
                 token = c;
 
-                const int color = optionsParseRequireRange(
-                        optionsParseRequiredNumber(c), 0 ,255);
+                const int color = options_parsenum(c, 0, 255, &errmsg);
```
daltomi, line 172: The last color digit is not being parsed, when using the following: `...-c 0,255,0,150 -t 'Hi'"` you get an error:
```bash
scrot: option --note: color '150 -t 'Hi'' is not a number
```
- The problem is in function `options_parsenum`, when it detects if it is a number or not. The solution would be what the manpage of `strtoll` indicates.

```diff
-    } else if (*end != '\0') {                                                                  
+    } else if (str == end) {         
```

- The other solution is to replace the space character just before calling the function:
```diff
+                char *const space = strchr(c, ' ');
+
+                if (space) *space = '\0';
+
                 const int color = options_parsenum(c, 0, 255, &errmsg);
                 if (errmsg) {
                         errx(EXIT_FAILURE, "option --note: color '%s' is %s", c,
                             errmsg);
                 }
 
+                if (space) *space = ' ';
```

---
I prefer this second solution because with `if (*end != '\0') ` it does not allow for example a string like `22a`, I think it is better.

guijan: That will have unwanted fallout, if you pass a color like "1 2,3,4" the error message won't look quite right. The right solution would probably use strspn() to skip over all the numbers and commas and check if a space follows, I'll have a look at it later.

daltomi: I don't understand what error you say. Your last commit 4c318bf also allows the form '1 2,3,4' without showing any errors.
And have 2 problems: first it doesn't allow the color to be used as the last parameter: `-t 'Hi' -c 255` and second with: `-c a,255`  say: `scrot: option --note: suboption -c: red channel '' is the null string`
but with my patch it works correctly :` scrot: option --note: color 'a' is not a number` and it does not change the behavior of the option `--note` that allows to place the color at the end.

daltomi: The truth is that I do not like https://github.com/resurrecting-open-source-projects/scrot/commit/4c318bfa63b461b220c8580b84cf4f107174996a.
It's not "fair" for a function to change so much because of behavior outside of it, which was working fine.

For the problem that resulted in changing `optionsParseRequireRange` to `options_parsenum` my 3 lines of code are enough for now.
If a lot needs to be corrected, a new PR will open soon after I merge this one.

**I would really like this PR to be in the next `release` that will be released in the next few days** :(

guijan: I've squashed the commits and done the requested changes.

daltomi:
> Code style, name of functions: https://webkit.org/code-style-guidelines/#names
> 
> > Lower-case the first letter, including all letters in an acronym, in a variable or function name.

Ok, the last thing that would be missing is the style of the function name.

daltomi (changes requested):
>  Code style, name of functions: https://webkit.org/code-style-guidelines/#names
   >
   >     Lower-case the first letter, including all letters in an acronym, in a variable or function name.

Ok, the last thing that would be missing is the style of the function name.

guijan:
Can you suggest a name? I don't quite understand what the WebKit style wants here.
Should it be "OptionsParsenum"?

daltomi:
> Can you suggest a name? I don't quite understand what the WebKit style wants here. Should it be "OptionsParsenum"?

Should be `optionsParseNum()`

Btw, you can use sed:

```bash
sed 's/options_parsenum/optionsParseNum/g' -i src/{*.c,*.h}
```

guijan:
Changes made and commit squashed, thanks for the testing and suggestions and sorry for the delays.

daltomi (approved):

Issue #208, opened 2023-01-10
Can we avoid libbsd?


targzeta:
On my system I don't have the libbsd installed:
```
checking for libbsd-overlay... no
configure: error: BSD functions not found, libbsd is required
```

I have seen that we have the `--without-libbsd` option, but it doesn't seem to work:
`configure: error: BSD functions not found and --without-libbsd was used`

Is it possible avoiding the use of libbsd?

Thank you in advance,
Emanuele

guijan:
It's not. That option is there for feature testing the need for it. Perhaps it should be renamed.

N-R-K:
I recently realized (due to commit https://github.com/resurrecting-open-source-projects/scrot/commit/95efe09e18801b3e28e6708c364a5a7dddbe6fbb) that on glibc systems, libbsd is no longer a dependency in `master` branch (although this ended up happening due to coincidence).

Unfortunately, musl [doesn't provide `sys/queue.h`](https://wiki.musl-libc.org/faq.html). So on musl systems, you'll still need libbsd.

Looking through the issue tracker, I also see that a couple other people were unhappy about libbsd dependency too. But I don't have any plans to remove/rewrite `sys/queue.h` dependency as of now (although I didn't take a close look at what it's used for).

There's also no guarantee that glibc will stay libbsd free either, but if there's no compelling reason to (re)introduce hard dependency on libbsd (on glibc systems), then I'll probably keep it that way.

guijan:
Haiku, Illumos, and dietlibc don't have that header either. Last time I heard, Haiku was introducing an X11 compatibility layer to their original GUI toolkit, so maybe scrot works there these days. Reimplementing a dependency isn't any better than using the dependency, so even if someone removes the rest of the libbsd bits we use by writing alternatives, I'd really be against it. I intend to at minimum support all the currently maintained libre systems with a POSIX C API and any proprietary unices we can get access to, we've made great progress in that direction during the last 2 years (scrot barely built outside of Debian without patching before I arrived), and libbsd's interfaces help with doing so without confining scrot to the ugliest subset of the POSIX APIs that everyone supports.

I'd expand its usage if I found a good place to use it, that's why I wrote as much in the TODO.md. `strsep()` could replace `strtok()`. `strsep()` is easier to use, but I haven't paid attention to the `strtok()` calls in scrot's source to see if they run into `strtok()`'s weirdness.

PR #209, opened 2023-01-10
fix libbsd feature test confusion

guijan:
Users expressed confusion when using the old "--without-libbsd" configure flag. Some assumed that it compiles scrot without libbsd, but it only made the build script error if libbsd is a dependency. This is useful to package maintainers: it gives them a yes/no answer on whether the library is a dependency on their system.

The old configure flag has been replaced by a more verbose "--enable-libbsd-feature-test".

See: https://github.com/resurrecting-open-source-projects/scrot/issues/208 https://github.com/resurrecting-open-source-projects/scrot/discussions/187

I've tested the yes/no paths using Alpine Linux and OpenBSD.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ scrot requires a few projects and libraries:
 - [autoconf-archive](https://www.gnu.org/software/autoconf-archive/) (build time only)
 - [pkg-config](https://www.freedesktop.org/wiki/Software/pkg-config/) (build time only)
 - [imlib2](https://sourceforge.net/projects/enlightenment/files/imlib2-src/)
-- [libbsd](https://libbsd.freedesktop.org/wiki/) (if `./configure --enable-libbsd-feature-testing` returns true)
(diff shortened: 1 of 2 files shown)


daltomi:
Let's hope no one reads fast and doesn't see the `-test` at the end, because it also looks like you're enabling the libbsd library features: "If I enable libbsd it will be more secure"(or something like that)
But if you prefer as it is we leave it like that.

eribertomota:
I will wait for this change to release the 1.8.1 version.

daltomi:
> But if you prefer as it is we leave it like that.

Hi @guijan
Please, if you can, confirm that you do not want to make changes to this PR, thank you.

guijan:
Hi, yes, the PR is finished, I don't intend to change it.

daltomi:
@eribertomota all ready to merge then, FYI this PR should close at #208 (Can we avoid libbsd?), thanks you.

eribertomota:
> @eribertomota all ready to merge then, FYI this PR should close at #208 (Can we avoid libbsd?), thanks you.

Thanks, @daltomi.

PR #210, opened 2023-01-13
Improve make-dist

eribertomota:
* Add some files to bootstrapped tarballs.
* Improve build and install instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,12 @@ make install
 
 Cleaning up the build is also simple:
 ```
-$ make distclean
+make clean
+```
+
+To vanish the source code, removing all pre-built files, including configure, run:
+```
+make distclean
 ```
 
 You can return to a pristine source tree before running `./configure`:
--- a/README.md
+++ b/README.md
@@ -35,6 +35,10 @@ If you are interested in helping scrot, read the [CONTRIBUTING.md](CONTRIBUTING.
 
 ## Building ##
 
+This section describes the steps to build and install scrot.
+
+### Dependencies ###
+
 scrot requires a few projects and libraries:
 
 - [autoconf](https://www.gnu.org/software/autoconf/autoconf.html) (build time only)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -48,28 +52,40 @@ scrot requires a few projects and libraries:
...
+And as root:
 ```
-$ ./autogen.sh clean
+make install
 ```
 
-## Installing ##
+Cleaning up the build is also simple:
+```
+$ make distclean
```
daltomi, line 83: Note that `make clean` only cleans the build(`rm src/scrot; rm src/*.o`) but `distclean` removes all build + configure script: the user is forced to run `autogen.sh` in the latter case.

eribertomota: Yes, right. What you think about:


Cleaning up the build is also simple:
```
make clean
```

To vanish the source code, removing all pre-built files, including ```configure```, run:
```
make distclean
```

daltomi: Yes, I agree

daltomi (approved):

README.md
daltomi: > Yes, right. What you think about:
> 
> Cleaning up the build is also simple:
> 
> ```
> make clean
> ```
> 
> To vanish the source code, removing all pre-built files, including `configure`, run:
> 
> ```
> make distclean
> ```


My web browser went crazy, it doesn't publish some comments or they are duplicated xD
But just in case, to this your query, is `Yes, I agree`

eribertomota:
Only to confirm: @daltomi, could I approve this PR now as is?

eribertomota:
I just updated the PR.

daltomi:
> Only to confirm: @daltomi, could I approve this PR now as is?

Yes, it's fine for me.

README.md
```diff
@@ -48,28 +52,45 @@ scrot requires a few projects and libraries:
...
+And as root:
 ```
-$ ./autogen.sh clean
+make install
 ```
 
-## Installing ##
+Cleaning up the build is also simple:
+```
+make clean
```
daltomi, line 83: Wait..
Missing the `$` character:`$ make clean` and `# make install`

daltomi (changes requested):

README.md
eribertomota: I removed because all other  previous commands was without the prompts. I prefer all commands with prompts. What you think?

daltomi: `$ ./autogen.sh` 
`$ ./configure`
 `$ make` 
`$ make distclean`
and others that must be out there. :)

daltomi: > I removed because all other previous commands was without the prompts. I prefer all commands with prompts. What you think?

Yes, I agree,  I also prefer it with prompts.

daltomi (approved):

eribertomota:
Thanks @daltomi

PR #212, opened 2023-01-18
New option: --window

daltomi:
```
It allows to indicate the identifier of the window, value in
hexadecimal, to which to capture.

Ex: 'scrot --window 0x100000c'
```

Ex:
```fish
 set WID $(xwininfo | grep "Window id:" | awk '{print $4}') && scrot --window $WID
```

Required parent branch: 23d587a ("Add delay to all capture options")

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -85,16 +85,26 @@ struct ScrotOptions opt = {
     .lineMode = LINE_MODE_S_CLASSIC,
     .stackDirection = HORIZONTAL,
     .monitor = -1,
+    .windowId = None,
 };
 
 static void showUsage(void);
 static void showVersion(void);
+static long long optionsParseNumBase(const char *, long long, long long,
+    const char *[static 1], int);
 static void optionsParseThumbnail(char *);
 
-/* optionsParseNum: "string to number" function.
+long long optionsParseNum(const char *str, long long min, long long max,
+    const char *errmsg[static 1])
+{
+    return optionsParseNumBase(str, min, max, errmsg, 10);
+}
+
+/* optionsParseNumBase: "string to number" function.
  *
  * Parses the string representation of an integer in str, and simultaneously
- * ensures that it is >= min and <= max.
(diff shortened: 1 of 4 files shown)


guijan:
The feature is fine, but is there a preexisting need for it?

Also, I wouldn't allow hex numbers in every option just because the C interface makes it easy. That's exposing an implementation detail rather than an actual design choice.
There should be an `optionsParseNumBase()` function that is `optionsParseNum()` but with an extra parameter that is the base `strtoll()` uses. `optionsParseNum()`'s usage and behavior would remain unchanged, but its implementation would then become a call to `optionsParseNumBase()` with the base set to 10.

daltomi:
I like the idea of a function like this, could you send a PR to this same branch with the changes? So your authorship would also remain in this PR.

guijan (commented):
The call to OptionsParseNumBase should use 0 as the argument to the `base` parameter.
xdotool returns a base 10 integer.
xwininfo returns a base 16 integer by default, but you can also tell it to return base 10.

daltomi:
> The call to OptionsParseNumBase should use 0 as the argument to the `base` parameter. xdotool returns a base 10 integer. xwininfo returns a base 16 integer by default, but you can also tell it to return base 10.

Yes, you're right, I'll change it now

src/options.c
```diff
@@ -469,6 +480,13 @@ void optionsParse(int argc, char *argv[])
                     errmsg);
             }
             break;
+        case 'w':
+            opt.windowId = optionsParseNumBase(optarg, None/*0L*/, LONG_MAX, &errmsg, 0);
```
N-R-K, line 484: One concern I have is that it might be valid to have window id beyond `LONG_MAX` since `Window` is typically typedef to `unsigned long`. But I'm not sure about that.

N-R-K (approved):
Looks OK and works OK for my testing too.

I've squashed the commits (while preserving author and giving co-author credits) and plan on merging this.

PR #214, opened 2023-01-21
Indicate that the value is not an integer

daltomi:
Until version v1.7 it would silently convert a floating point number to a zero(0). Which is an error since it allowed the user to believe that the value given to the option was correct, when it is not. Since version v1.8, the entered value is indicated with an error 'is not an integer'.

### DIFF
--- a/FAQ
+++ b/FAQ
@@ -149,4 +149,4 @@
 
    For example:
    $ scrot --delay 0.5
-   $ scrot: option --delay: '0.5' is is not an integer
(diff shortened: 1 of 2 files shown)


guijan:
We can just add support for it instead.

daltomi:
I don't see any use for it.
The behavior of this option is in seconds and it stays that way.
Some users probably read somewhere that a precision delay (never worked) was needed to fix other issues, maybe the "resource not available" error when recording the keyboard or something like that.

PR #217, opened 2023-01-26
restore "keep ratio" functionality in --thumb

### LINKED ISSUE #216: keep aspect ratio for thumb
I use to run this command to capture the screen, including a thumb:
```
scrot --thumb 100x0 /tmp/screen.png
```

The thumb used to have a width of 100 pixels, preserving the correct aspect ration. This broke with ff88b0b0cd38137113319cb86e423cb1fe4d6b3a (I think, to be verified...) and returns an error message now:
```
scrot: option --thumb: resolution height '0' is too small
```

Can we define a special case for value `0` and have the old behavior back?

guijan:
Fixes #216 and documents this previously undocumented feature.

### DIFF
--- a/man/scrot.txt
+++ b/man/scrot.txt
@@ -4,7 +4,7 @@ NAME
 SYNOPSIS
   scrot [-bcfhimopuvz] [-a X,Y,W,H] [-C NAME] [-D DISPLAY] [-d SEC] [-e CMD]
         [-F FILE] [-k OPT] [-l STYLE] [-M NUM] [-n OPTS] [-q NUM] [-S CMD] [-s OPTS]
-        [-t NUM | GEOM] [FILE]
+        [-t % | WxH] [FILE]
 
 DESCRIPTION
   scrot (SCReenshOT) is a simple command line screen capture utility, it uses
@@ -56,10 +56,15 @@ OPTIONS
   -s, --select OPTS         Interactively select a window or rectangle with the
                             mouse, use the arrow keys to resize. See the -l and
                             -f options. OPTS it's optional; see SELECTION MODE
-  -t, --thumb NUM | GEOM    Also generate a thumbnail. The argument is the
-                            resolution of the thumbnail, it may be a percentage
-                            NUM or a resolution GEOM. Examples: 10, 25, 320x240,
-                            500x200.
+  -t, --thumb % | WxH       Also generate a thumbnail. The argument represents
+                            the thumbnail's resolution: if the argument is a
+                            single number, it is a percentage of the full size
+                            screenshot's resolution; if it is 2 numbers
+                            separated by an "x" character, it is a resolution.
+                            If one of the resolution's dimensions is 0, it is
+                            replaced by a number that maintains the full size
+                            screenshot's aspect ratio. Examples: 10, 25, 320x240,
+                            500x200, 100x0, 0x480.
   -u, --focused             Use the currently focused window.
   -v, --version             Output version information and exit.
   -z, --silent              Prevent beeping.
--- a/src/options.c
+++ b/src/options.c
@@ -509,7 +509,7 @@ void optionsParse(int argc, char *argv[])
                 free(opt.outputFile);
                 opt.outputFile = getPathOfStdout();
                 opt.overwrite = 1;
-                opt.thumbWorP = 0;
+                opt.thumb = THUMB_DISABLED;
             }
         } else
             warnx("unrecognised option %s", argv[optind++]);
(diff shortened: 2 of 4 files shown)


daltomi:
This works: `scrot -t 0x1`

But not the following: `scrot -t 1x0`
`scrot: unable to create thumbnail: No such file or directory`

guijan:
Hmm, that bug is also in ae4deba4772f543842933184722aeb77b49496cd.
It's probably something to do with the height being rounded to 0 and something elsewhere erroring. Need to investigate.

guijan:
Scrot really needs automated testing, all that testing you do needs to be a few keypresses away...

man/scrot.txt
```diff
@@ -4,7 +4,7 @@ NAME
 SYNOPSIS
   scrot [-bcfhimopuvz] [-a X,Y,W,H] [-C NAME] [-D DISPLAY] [-d SEC] [-e CMD]
         [-F FILE] [-k OPT] [-l STYLE] [-M NUM] [-n OPTS] [-q NUM] [-S CMD] [-s OPTS]
-        [-t NUM | GEOM] [FILE]
+        [-t PCT | WxH] [FILE]
```
N-R-K, line 7: `PCT` is very awkward. I'd either spell it out fully or use `%`.

src/options.c
```diff
@@ -85,6 +85,8 @@ struct ScrotOptions opt = {
     .lineMode = LINE_MODE_S_CLASSIC,
     .stackDirection = HORIZONTAL,
     .monitor = -1,
+    .thumbWorP = -1,
```
N-R-K, line 88: Mostly a style nit: but I'd prefer two different variables instead of overloading one.

Feel free to ignore if doing so is too much hassle.

N-R-K (changes requested):

man/scrot.txt
guijan: % seems fine to me, I will change it.

src/options.c
guijan: I think it's a good idea that the code should be obvious instead of unobvious and commented, I'll write a version that uses a separate variable and push it if it's simpler.

N-R-K: Any update on this? This can be done a later time, it's not as important as restoring previous functionality.

Also I'm thinking that the style related change should be done on a separate commit (or PR).

guijan:
I've tested the following invocations:
```sh
src/scrot --thumb 0 /tmp/test.png
src/scrot --thumb 0x0 /tmp/test.png
src/scrot --thumb 0x1 /tmp/test.png
src/scrot --thumb 1x0 /tmp/test.png
src/scrot --thumb 1x1 /tmp/test.png
src/scrot --thumb 2x0 /tmp/test.png
src/scrot --thumb 2x1 /tmp/test.png
src/scrot --thumb x100 /tmp/test.png
src/scrot --thumb 50 /tmp/test.png
```
They all produced the appropriate output files or error messages.

src/options.h
```diff
@@ -54,11 +54,13 @@ struct ScrotOptions {
...
-    /* If != -1, the --thumb option is in use. */
-    int thumbWorP;
-    /* If thumbH != -1, thumbWorP contains a width and thumbH contains a
-     * height. Otherwise, thumbWorP contains a percentage.
+    /* If 0, not in use.
+     * If -1, use thumbW and thumbH as the thumbnail's resolution.
+     * If between 1 inclusive and 100 inclusive, the thumbnail's resolution is
+     * opt.thumb% of the original's resolution.
      */
+    int thumb;
```
N-R-K, line 62: Since goal was to stop overloading a single variable with multiple meaning, I think the following would be more useful:

```c
enum { THUMB_DISABLED, THUMB_PERCENT, THUMB_RES } thumb;
int thumbPercent;
```

With this, I don't think there needs to be any comment in the code at all, since it will be self-explanatory.

N-R-K (changes requested):
One nitpick on overloading `.thumb` but otherwise, it looks OK (will test it out soon-ish just in case).

N-R-K (approved):
Looks good and seems to be working OK from my testing.

One last thing, could you squash the last two commits and give it a better commit message. Something like "use descriptive constants so the code is obvious".

P.S: Was going to do the squashing myself, but then I noticed that you gpg sign your commits. So I didn't want to mess with it.

guijan:
I squashed all the commits in this branch instead.

N-R-K:
> I squashed all the commits in this branch instead.

Hmm, I'm on the fence about this. These type of commits where it says in the commit message "fix X" but the diff does more than that is a massive annoyance for future maintainers who might want to understand why a particular line was added/changed via `git blame`.

Ideally, it should be 3 different commits:

* Restore keep ratio
* Fix err to errx
* Refactor with named constants

But I'd be OK with it being squashed to one as long as the commit message's body mentions the reasons behind the `errx` and refactor change as well.

N-R-K:
Thanks.

Issue #215, opened 2023-01-26
Multiple bugs in scrot


guijan:
This is a general issue for tracking bugs I've found but not put in the work to fix before I forget them.

# Currently unfixed
## access()
The `access()` function suffers from a [time of check to time of use](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) issue. Nothing guarantees us that what `access()` could access will be there when we try to `open()`, or that what it couldn't access won't be there. The code shouldn't ask if it can do something, it should just do it and check the return value. This function shouldn't exist and its callers should be reviewed:
https://github.com/resurrecting-open-source-projects/scrot/blob/70a7540cfdc1000db8d400883c2eece386b5bf91/src/options.c#L332-L336
`access()` is also used elsewhere, use grep to find the uses:
https://github.com/resurrecting-open-source-projects/scrot/blob/70a7540cfdc1000db8d400883c2eece386b5bf91/src/scrot.c#L486
https://github.com/resurrecting-open-source-projects/scrot/blob/70a7540cfdc1000db8d400883c2eece386b5bf91/src/scrot.c#L455

## `scrot --delay` time drift
The implementation of the --delay option calls `sleep()` in a loop, time passes between each `sleep()` call, so it drifts. 
https://github.com/resurrecting-open-source-projects/scrot/blob/70a7540cfdc1000db8d400883c2eece386b5bf91/src/scrot.c#L262-L281
Edit: setitimer() fixes this but is extremely difficult to use. clock_nanosleep() also fixes it but is not widely implemented. Most timer APIs in POSIX are broken in some way or another, and the fixed versions are not widely implemented, so this will have to wait. There was an attempt at #218.

## Maximum filename length is wrong
Scrot uses this macro to check if input/output filenames are within a certain size:
https://github.com/resurrecting-open-source-projects/scrot/blob/70a7540cfdc1000db8d400883c2eece386b5bf91/src/options.c#L61-L66
https://github.com/resurrecting-open-source-projects/scrot/blob/70a7540cfdc1000db8d400883c2eece386b5bf91/src/options.c#L76

The issue is that a maximum length of 256 is hardcoded, but the maximum filename length is a very complex issue. 256 could be too much or too little depending on the operating system, the file system, and even what directory the file is created in. The solution is that the code handling input/output filenames should pass around an alloc storage duration buffer and dynamically grow it as needed throughout the program's execution. Then, when the file is finally created, the system will return error if it fails to create the file.

[...]

Issue #213, opened 2023-01-21
non-integer delay is not accepted


a1ea321:
`scrot -d 0.5` stopped working. It spits out: `scrot: option --delay: '0.5' is not a number`

I believe this is after I updated scrot through the package manager in Artix.

I `git clone`d and compiled scrot (tag:1.8.1). The problem is there. Then I tried the scrot obtained through `sudo apt install scrot`. That one works.

If you have nothing against restoring this functionality, please do it or say "do it" to me.

daltomi:
Hi @a1ea321 

Yeah, that doesn't look good; it's confusing.
I don't know where I've seen floating point used before but a value like that is just 0.

The improvement we could make is to indicate, via manpage or error message, that `scrot `does not use floating point numbers in any of its options.

- A note for collaborators in general:
Since early versions, eg `scrot v0.8`, a value of `0.5` results in `0`; internally `atoi()` and `sleep()` do not support floating point numbers;  of course, the difference is that `v0.8` quietly allowed this.

a1ea321:
So you're telling me... All this time I've been using `scrot -d 0.5` but the delay did not have any effect x-D Not my first time cargo cult programming.

I added this for a full screen screenshot triggered by dmenu based command launcher. I vaguely remember the 0.5 being necessary for dmenu to get out of the way. (1 second would have been too long of a wait.)

But now, I can't reproduce that issue. Zero delay seems enough time for dmenu. 

Thanks!

PR #218, opened 2023-02-01
fix scrot --delay drift

guijan:
This PR still needs testing, but it's otherwise complete.

It has a workaround for an OpenBSD bug, someone has to check if this bug is present in other operating systems. I can probably bake the logic to check for the bug into the code, but that will have to wait.

Edit: I thought it was possible the other BSDs suffered from the same issue (siginfo_t unimplemented) because they share the same origin, but it's only OpenBSD, so I used a small #ifdef instead of writing the logic to check for the bug

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -36,12 +36,12 @@ bin_PROGRAMS  = scrot
 
 scrot_CPPFLAGS = @CPPFLAGS@
 scrot_LDADD    = @LIBS@
-scrot_SOURCES  = scrot.c scrot.h        \
-imlib.c imlib.h                         \
-note.c note.h                           \
-options.c options.h                     \
-scrot_selection.c scrot_selection.h     \
-selection_classic.c selection_classic.h \
-selection_edge.c selection_edge.h       \
-slist.h                                 \
+scrot_SOURCES  = scrot.c scrot.h                        \
+imlib.c imlib.h                                         \
+note.c note.h                                           \
+options.c options.h options_delay.c options_delay.h     \
+scrot_selection.c scrot_selection.h                     \
+selection_classic.c selection_classic.h                 \
+selection_edge.c selection_edge.h                       \
+slist.h                                                 \
 util.c util.h
--- a/src/options.c
+++ b/src/options.c
@@ -55,6 +55,7 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 #include "note.h"
 #include "options.h"
 #include "scrot.h"
+#include "options_delay.h"
 #include "scrot_selection.h"
 #include "util.h"
 
@@ -415,6 +416,7 @@ void optionsParse(int argc, char *argv[])
                 errx(EXIT_FAILURE, "option --delay: '%s' is %s", optarg,
                     errmsg);
             }
+            scrotTimerStart();
             break;
         case 'e':
             opt.exec = estrdup(optarg);
--- a/src/options_delay.c
+++ b/src/options_delay.c
@@ -1,0 +1,181 @@
+/*
+
+Copyright 2023 Guilherme Janczak <guilherme.janczak@yandex.com>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to
+deal in the Software without restriction, including without limitation the
+rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+sell copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies of the Software and its documentation and acknowledgment shall be
+given in the documentation and software packages that this Software was
+used.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+*/
+
+#include <sys/time.h>
+
+#include <err.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <poll.h>
+#include <signal.h>
+#include <stdio.h>
+#include <unistd.h>
+
+#include "options.h"
+
+/* Initialized to invalid fds. */
+static int selfPipe[] = {-1, -1};
+
+static void sigalrm(int, siginfo_t *, void *);
+static void timerEnd(void);
+
+/* scrotTimerStart: start counting seconds. */
+void
+scrotTimerStart(void)
+{
+    struct sigaction act;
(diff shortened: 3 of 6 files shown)


guijan:
There's the minor detail that if you combine a delay and selecting a rectangular area, this PR changes picking the selection area happen after the delay, not before as was previously the case. The fact that the code used to go out of its way to make the old behavior happen was really weird to me and so was the code itself. Let me know if this change isn't okay.

daltomi:
El Thu, 02 Feb 2023 19:32:48 -0800
Guilherme Janczak ***@***.***> escribió:
> There's the minor detail that if you combine a delay and selecting a rectangular area, this PR changes picking the selection area happen after the delay, not before as was previously the case. The fact that the code used to go out of its way to make the old behavior happen was really weird to me and so was the code itself. Let me know if this change isn't okay.
> 

Look at number 2 of the FAQ. This behavior should stay.

PS: Lately I am very busy, I may not be able to participate much in the project, but as soon as I can help, I will.

N-R-K:
Can you elaborate a bit on what exactly is being fixed here (and some steps to reproduce)?

If it's just trying to fix `sleep` waking up early, then this seems insanely over-complicated for that task. You could instead use `nanosleep` - which returns the remaining time in case of an early wake up via the `rem` argument.

N-R-K:
Okay, I just saw https://github.com/resurrecting-open-source-projects/scrot/issues/215.

> The implementation of the --delay option calls `sleep()` in a loop, time passes between each `sleep()` call, so it drifts.

First of all, scrot only accepts a delay at ***seconds*** granularity (see https://github.com/resurrecting-open-source-projects/scrot/issues/213). Secondly, this isn't a video game where missing the deadline by a couple milisec would result in a potential bad user experience (i.e frame-drops).

Given these information, it's not practical to add over-complicated code for this IMHO.

A more practical approach (that doesn't require hundreds of line of code) would be to:

* Use a wrapper around `nanosleep` that loops in case of early wake up.
* Use `dprintf` to bypass `stdio` buffering.

This is still not going to be 100% accurate - but once again - we're dealing with granularity of *seconds* here. Missing the deadline by a couple milisec is really not a big deal for scrot's usercase IMO.

guijan:
>missing the deadline by a couple milisec
There's no guarantee that only "a couple milisec" pass between each sleep() call. 

>Can you elaborate a bit on what exactly is being fixed here (and some steps to reproduce) ?
Currently, it fixes the drift calling sleep() in a loop causes, and it also also fixes any stray SIGALRM reducing scrot's delay which I noticed while fixing the former.

Call scrot with a long delay and a countdown in one terminal:
```console
$ scrot -cd 99999
```
Then send it SIGALRM in a loop from another:
```console
$ while true; do pkill -SIGALRM scrot; done
```
You will see scrot count down really fast.

This PR also starts the delay as soon as we know -d was used which is nice for greater accuracy.

Also, I just noticed yet another bug in both this PR and the current scrot master: stdout could block.

N-R-K:
> You will see scrot master count down really fast.

I mentioned this in my post - you can fix this with `nanosleep`. Here's a patch:

```diff
diff --git a/src/scrot.c b/src/scrot.c
index 295a21f..4dc66d1 100644
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -259,24 +259,28 @@ static Imlib_Image scrotGrabAutoselect(void)
     return im;
 }
 
+static void
+xsleep(int sec)
+{
+    struct timespec delay = { .tv_sec = sec };
+    while (nanosleep(&delay, &delay) < 0 && errno == EINTR);
+}
+
 void scrotDoDelay(void)
 {
     if (opt.delay) {
         if (opt.countdown) {
             int i;
 
-            printf("Taking shot in %d.. ", opt.delay);
-            fflush(stdout);
-            sleep(1);
+            dprintf(1, "Taking shot in %d.. ", opt.delay);
+            xsleep(1);
             for (i = opt.delay - 1; i > 0; i--) {
-                printf("%d.. ", i);
-                fflush(stdout);
-                sleep(1);
+                dprintf(1, "%d.. ", i);
+                xsleep(1);
             }
-            printf("0.\n");
-            fflush(stdout);
+            dprintf(1, "0.\n");
         } else
-            sleep(opt.delay);
+            xsleep(opt.delay);
     }
 }
```

> starts the delay as soon as we know -d was used which is nice for greater accuracy.

It'll still miss the time it took to get to that point - which can be noticeable if you try to load the binary (the first time) from a really slow HDD (or NFS).

> It's also fixable.

Not entirely, there's still going to be inaccuracies due to clock resolution (and the startup time I mentioned above). But even if we assume it's fully fixable - that's besides the point.

My point was that this is adding hundreds of lines of code into a project - so the question of "What benefit the end-user get out of this?" is worth asking. And I don't see how this is going to be beneficial to 99.99999% of the user-base.

In short, I don't think the amount of code added vs benefit provided ratio is anywhere near being worth it.

But that's just my opinion, of course :) If the project maintainer thinks otherwise then go for it.

guijan:
There are timing issues we can't fix like the time the processor takes to do things or the time between the user pressing enter and main() starting, yes. Those aren't scrot's fault. I'm fixing the ones that are scrot's fault.

The code's semantics in master allow for -d to sleep for an indeterminate amount of time longer than what was requested. Probably, the OS and machine will respond quickly enough. By the same token, access() probably checked the same file as the one we are open()ing later, malloc() probably won't error, stdout probably won't block and cause us to wait even longer than requested, but those are all bugs. I'd rather have a few hundred extra LoC that do the right thing (even if they may have bugs of their own) than to keep the bug.

N-R-K:
> By the same token ...

It depends on the cost. Some of these (like `malloc`) can be fixed with low/no cost.

> I'd rather have a few hundred extra LoC that do the right thing (even if they may have bugs of their own) than to keep the bug.

I believe that's not a pragmatic mindset and does more harm than good in the long run. So we'll just have to agree to disagree on this.

guijan:
Currently writing an alternative solution that uses `clock_nanosleep()` if available, `pthread_cond_timedwait()` otherwise. It's less LoC and no signal handler.

N-R-K:
> Currently writing an alternative solution that uses `clock_nanosleep()` if available, `pthread_cond_timedwait()` otherwise. It's less LoC and no signal handler.

Adding threads for this is still far into overkill territory, IMO. 

> Indeed, it seems the current solution is the one with the correct semantics that runs on the most platforms.

On my system (Linux 6.1.9 & glibc 2.36), trying to use any delay just hangs indefinitely. Attaching `gdb` into the hanged process shows that it's stuck on `poll`.

N-R-K:
> This PR also starts the delay as soon as we know -d was used which is nice for greater accuracy.

I think this breaks the `-s` behavior. The delay should start _after_ the selection has been made, not before.

guijan:
> I think this breaks the `-s` behavior. The delay should start _after_ the selection has been made, not before.

I noticed that, I thought the time the user takes to create a selection being added to the delay was an error.

>trying to use any delay just hangs indefinitely. 

One more thing to investigate...
Writing this, I learned that most of the POSIX timer APIs are broken and that the APIs that aren't broken are rarely implemented.

guijan:
Better to just close the PR, and try again when the fixed POSIX APIs are more widely implemented.

N-R-K:
> On my system (Linux 6.1.9 & glibc 2.36), trying to use any delay just hangs indefinitely. Attaching `gdb` into the hanged process shows that it's stuck on `poll`.

I'll just leave it out here for archival purposes: The reason for it seems to be this check:

```c
    if (info->si_code != SI_TIMER)
        return; /* Someone gave us a SIGALRM. Ignore it. */
```

It was always returning early and so nothing was getting written to the pipe.

Gdb print of `*info`

```c
{si_signo = 14, si_errno = 0, si_code = 128, /* everything else was 0... */
```

And after grepping around a bit in `/usr/include/`, 128 seems to be SI_KERNEL. (Usually I wouldn't go looking inside headers and would rather read documentation, but I'm not too interested in this, so this is more of an info-dump).

- - -

Also according to the linux-manpage, `setitimer` seems to be declared obsolete by POSIX.

PR #220, opened 2023-02-13
do not hardcode string length

N-R-K:
just declare an array and use sizeof operator on it. less error prone this way.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -539,7 +539,7 @@ char *optionsNameThumbnail(const char *name)
 {
     const char thumbSuffix[] = "-thumb";
     const size_t newNameLength = strlen(name) + sizeof(thumbSuffix);
-    char *newName = calloc(1, newNameLength);
+    char *newName = malloc(newNameLength);
 
     if (!newName)
         err(EXIT_FAILURE, "Unable to allocate thumbnail");
--- a/src/options.c
+++ b/src/options.c
@@ -18,6 +18,7 @@ Copyright 2021      IFo Hancroft <contact@ifohancroft.com>
 Copyright 2021      Peter Wu <peterwu@hotmail.com>
 Copyright 2021      Wilson Smith <01wsmith+gh@gmail.com>
 Copyright 2022      Zev Weiss <zev@bewilderbeest.net>
+Copyright 2023      NRK <nrk@disroot.org>
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to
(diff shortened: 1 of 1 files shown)


src/options.c
```diff
@@ -536,9 +537,8 @@ static void showVersion(void)
 
 char *optionsNameThumbnail(const char *name)
 {
-    const char *const thumbSuffix = "-thumb";
-    const size_t thumbSuffixLength = 7;
-    const size_t newNameLength = strlen(name) + thumbSuffixLength;
+    const char thumbSuffix[] = "-thumb";
+    const size_t newNameLength = strlen(name) + sizeof(thumbSuffix);
     char *newName = calloc(1, newNameLength);
```
guijan, line 542: Change this calloc call to malloc too.

guijan (changes requested):
Maybe this version of the whole function is better; it's definitely briefer, but I'm not sure if it's simpler:

```c
char *optionsNameThumbnail(const char *name)
{
    const char thumbSuffix[] = "-thumb";
    const size_t newNameLength = strlen(name) + sizeof(thumbSuffix);
    char *newName = malloc(newNameLength);

    if (!newName)
        err(EXIT_FAILURE, "Unable to allocate thumbnail");

    size_t basenameEnd = strcspn(name, ".");
    memcpy(newName, name, basenameEnd);
    strlcpy(newName+basenameEnd, thumbSuffix, newNameLength-basenameEnd);
    strlcat(newName, name+basenameEnd, newNameLength);

    return newName;
}
```
Up to you.

N-R-K:
> Change this calloc call to malloc too.

Sure, seems fine.

>     size_t basenameEnd = strcspn(name, ".");

I don't believe this would work, e.g `a.b.c` - the extension here should be `.c` but this will end up selecting `.b.c` instead. `memrchr` is what you want here (widely supported, but unfortunately non-standard).

>     memcpy(newName, name, basenameEnd);
>     strlcpy(newName+basenameEnd, thumbSuffix, newNameLength-basenameEnd);
>     strlcat(newName, name+basenameEnd, newNameLength);

If we're gonna use `memcpy` then I'd go all the way and use `mempcpy` (non-standard, but takes a single line to implement `return (char *)memcpy(dst, src, n) + n;`).

But I see that this project prefers using these BSD functions - so I'd rather not bikeshed :) (My opinion of `strl{cpy,cat}` is not positive).

guijan:
Good catch. I think the function shouldn't be rewritten now.

guijan (approved):

PR #222, opened 2023-02-14
Fix unchecked calloc calls

N-R-K:
This replaces unchecked `calloc` calls with `ecalloc` - a `calloc` wrapper that exits upon failure.


guijan (approved):
I've grepped for `calloc` in the source and checked that every call was replaced and that the usage of `calloc` did not change and the PR is good.

PR #224, opened 2023-03-01
Add scrotSleep and use dprintf

N-R-K:
### Introduce scrotSleep() which deals with EINTR

sleep() may end up waking up eariler than requested due to signals. so
use nanosleep() and deal with EINTR in case it happens.

### scrotDoDelay: use dprintf to bypass stdio stream

since we want the output to be immediately shown, using stdio doesn't
make much sense. using dprintf to output straight to the file descriptor
allows us to skip having to manually flush every time.

### scrotDoDelay: output countdown to stderr

otherwise, the countdown won't be visible when user is outputting the
image to stdout. for example commands like the following will end up
with a "corrupted" image due to the countdown being slipping in there.

    scrot -d3 -c - | xclip -in -selection clipboard -target image/png

by outputting the countdown to stderr, the above command will work
properly.

for reference, `maim` also outputs it's countdown to stderr, not stdout.

### DIFF
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -259,6 +259,17 @@ static Imlib_Image scrotGrabAutoselect(void)
     return im;
 }
 
+/* similar to sleep, but deals with EINTR.
+ * ideally, we'd want to sleep against an absolute time to prevent any drift
+ * but clock_nanosleep doesn't seem to be widely implemented.
+ */
+static void scrotSleep(int sec)
+{
+    assert(sec > 0);
+    struct timespec delay = { .tv_sec = sec };
+    while (nanosleep(&delay, &delay) < 0 && errno == EINTR);
+}
+
 void scrotDoDelay(void)
 {
     if (opt.delay) {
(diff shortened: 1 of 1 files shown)


guijan:
Add a comment mentioning that calling nanosleep() in a loop causes drift, but that it's hard to fix because clock_nanosleep() is not implemented everywhere and other relevant POSIX functions are either too difficult to use or also unimplemented.

Also, if PIPE_BUF runs out, dprintf() will block, assuming stdout is a pipe anyway.

N-R-K:
> Also, if PIPE_BUF runs out, dprintf() will block, assuming stdout is a pipe anyway.

Does scrot support outputting to stdout? If no, then I don't think PIPE_BUF will ever be reached. If yes, then we probably shouldn't be printing countdown to `stdout` anyways (`stderr` would be more suitable).

guijan:
You can definitely count down for long enough to run out of PIPE_BUF, it's 4096 bytes in most implementations and POSIX says it has to be at least 512 bytes. And we do support stdout output: https://github.com/resurrecting-open-source-projects/scrot/blob/4c59bd454f51401a22820d3a38973475da93cda4/man/scrot.1#L155-L157
But the countdown happens before the image is output of course. 

And yeah, the countdown should either be omitted or go to stderr if the output file is stdout.

Possibly, this PR could go through as is because at least it fixes one bug, if it fixes every single bug in this rather poor code is not too important, they can be addressed separately.

guijan (approved):

Issue #221, opened 2023-02-13
Screenshot delay with select option not working intuitively


kucharskim:
scrot first tries to select then does the delay, I think from user experience perspective, it should first wait and sleep for the delay and then allow user to select area for screenshot. In practice I often do `sleep 5; scrot -s` instead of `scrot -d5 -s`

kucharskim:
This was reported at the time on:
```
$ scrot -v 
scrot version 1.7
```

guijan:
Yes, I actually found the same thing weird while working on a PR. If we know what will be on the screen at a particular moment in the future, it's probably because it's on the screen now and we can screenshot it now. If we don't know what will be on the screen in the future, the current behavior is useless. It's also the case that when using scrot interactively, immediately showing a selection rectangle means the terminal that scrot is in is on the screen, possibly obscuring what is being screenshotted. I support this change.

There was the detail that the current behavior allows taking a screenshot of the mouse pointer while using a selection rectangle, but I don't think that's important, or that we should support drawing a rectangle both before and after the delay. It might also be fixable but I'm not familiar with GUI programming.

N-R-K:
The FAQ (number 2) mentions one https://github.com/resurrecting-open-source-projects/scrot/blob/master/FAQ

One other situation where it's useful, and I've faced this situation in practice, is when you're trying to take a screenshot of some pesky UI element that only shows itself when the mouse cursor is at a certain place.

Adding another flag to specify when the delay happens will take care of both use cases, I think. (Although the unix mindset in me thinks why bother when `sleep 5; scrot -s` works already).

guijan:
>take a screenshot of some pesky UI element that only shows itself when the mouse cursor is at a certain place.

Maybe combining -f with -s would fix that use case?

N-R-K:
> Maybe combining -f with -s would fix that use case?

I don't think it would. Switching away to a terminal will make the element disappear. And even if you bind it to a key - a lot of these elements disappear as soon as there's any keypress.

kucharskim:
> The FAQ (number 2) mentions one https://github.com/resurrecting-open-source-projects/scrot/blob/master/FAQ

Ah, make sense, good use case.

N-R-K:
I'm thinking of adding something like `--delay-select` which will delay the selection also (if active). Does anyone have ideas for a better long-opt name?

I'm also thinking of keeping this a long-opt only, unless someone can suggest a sensible short-opt that isn't taken already.

[...]

PR #227, opened 2023-03-17
getopt_long() cleanup

N-R-K:
Noticed these while I was trying to add `--delay-select` for #221 

### options: test against correct return value

getopt_long manpage says:

| If all command-line options have been parsed, then getopt() returns -1.

it doesn't say anything about returning EOF. but since EOF is typically
defined as -1 it ended up working by accident.

### options: remove unused variable

getopt_long accepts NULL as longindex argument, so no need to pass a
dummy variable.


guijan (approved):

N-R-K:
Thanks, I've changed the commit message.

PR #229, opened 2023-03-17
fix multiple bugs in output file handling

guijan:
Check the commit message for details.

This the extent of the testing I've done:
```console
$ src/scrot -F test test2
scrot: both the --file option and the FILE parameter were used
$ src/scrot -F test test2 test3
scrot: both the --file option and the FILE parameter were used
$ src/scrot test test2
scrot: extraneous non-option arguments: test2
$ src/scrot test test2 test3 test4
scrot: extraneous non-option arguments: test2 test3 test4
$ src/scrot "$(yes | tr -d '\n' | dd bs=251 count=1 2>/dev/null)".png
$ src/scrot "$(yes | tr -d '\n' | dd bs=256 count=1 2>/dev/null)".png
scrot: Saving to file yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy.png failed: File name too long
```
The last one shows that if we try to create a file with a name over NAME_MAX, imlib will return error without us having to do any checking which is exactly what we want. This PR increases the length limit of the output file's full path to fpathconf()'s idea of PATH_MAX or NAME_MAX, whichever comes first.

### DIFF
--- a/man/scrot.txt
+++ b/man/scrot.txt
@@ -3,8 +3,8 @@ NAME
 
 SYNOPSIS
   scrot [-bcfhimopuvz] [-a X,Y,W,H] [-C NAME] [-D DISPLAY] [-d SEC] [-e CMD]
-        [-F FILE] [-k OPT] [-l STYLE] [-M NUM] [-n OPTS] [-q NUM] [-S CMD] [-s OPTS]
-        [-t % | WxH] [-w NUM] [FILE]
+        [-k OPT] [-l STYLE] [-M NUM] [-n OPTS] [-q NUM] [-S CMD] [-s OPTS]
+        [-t % | WxH] [-w NUM] [[-F] FILE]
 
 DESCRIPTION
   scrot (SCReenshOT) is a simple command line screen capture utility, it uses
@@ -33,7 +33,6 @@ OPTIONS
                             When given the `b` prefix, e.g `-d b8`, the delay
                             will be applied before selection.
   -e, --exec CMD            Execute CMD on the saved image.
-  -F, --file                File name. See SPECIAL STRINGS.
   -f, --freeze              Freeze the screen when -s is used.
   -h, --help                Display help and exit.
   -i, --ignorekeyboard      Don't exit for keyboard input. ESC still exits.
@@ -72,8 +71,11 @@ OPTIONS
   -w, --window              Window identifier to capture.
                             Must be a valid identifier (see xwininfo(1)).
   -z, --silent              Prevent beeping.
-  -                         Redirection to standard output. The output image
-                            format is PNG.
+  -F, --file FILE           Specify the output file. If FILE is "-", scrot will
+                            output a PNG image to stdout. The filename is
+                            expanded according to the format specified in
+                            SPECIAL STRINGS. The output file may be specified
+                            through the -F option, or as a non-option argument.
 
 SPECIAL STRINGS
   -e, -F and FILE parameters can take format specifiers that are expanded
--- a/src/options.c
+++ b/src/options.c
@@ -66,9 +66,6 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
     }                                                           \
 } while(0)
 
-#define checkMaxOutputFileName(fileName) \
-    STR_LEN_MAX_FILENAME(output, (fileName))
-
 #define checkMaxInputFileName(fileName) \
     STR_LEN_MAX_FILENAME(input, (fileName))
 
@@ -86,6 +83,7 @@ struct ScrotOptions opt = {
     .stackDirection = HORIZONTAL,
     .monitor = -1,
     .windowId = None,
+    .outputFile = "%Y-%m-%d-%H%M%S_$wx$h_scrot.png",
 };
 
 static void showUsage(void);
@@ -93,6 +91,7 @@ static void showVersion(void);
 static long long optionsParseNumBase(const char *, long long, long long,
     const char *[static 1], int);
 static void optionsParseThumbnail(char *);
+static char *optionsNameThumbnail(const char *);
 
 long long optionsParseNum(const char *str, long long min, long long max,
     const char *errmsg[static 1])
@@ -386,6 +385,7 @@ void optionsParse(int argc, char *argv[])
     };
     int optch = 0;
     const char *errmsg;
+    bool FFlagSet = false;
 
     /* Now to pass some optionarinos */
     while ((optch = getopt_long(argc, argv, stropts, lopts, NULL)) != -1) {
@@ -475,7 +475,8 @@ void optionsParse(int argc, char *argv[])
             opt.script = estrdup(optarg);
             break;
         case 'F':
-            optionsParseFileName(optarg);
+            FFlagSet = true;
+            opt.outputFile = optarg;
             break;
         case 'M':
             opt.monitor = optionsParseNum(optarg, 0, INT_MAX, &errmsg);
@@ -497,35 +498,42 @@ void optionsParse(int argc, char *argv[])
             break;
         }
     }
+    argv += optind;
 
-    /* Now the leftovers, which must be files */
-    while (optind < argc) {
-        /* If recursive is NOT set, but the only argument is a directory
-           name, we grab all the files in there, but not subdirs */
-        if (!opt.outputFile) {
-            optionsParseFileName(argv[optind++]);
-
-            const bool redirectChar = ( opt.outputFile[0] == '-'
-                                        && opt.outputFile[1] == '\0');
-            if (redirectChar) {
-                optionsParseFileName(getPathOfStdout());
-                opt.overwrite = 1;
-                opt.thumb = THUMB_DISABLED;
-            }
-        } else
-            warnx("unrecognised option %s", argv[optind++]);
+    if (!FFlagSet && *argv) {
+        opt.outputFile = *argv;
+        argv++;
     }
-
-    /* So that we can safely be called again */
-    optind = 1;
+    if (*argv) {
+        /* If *argv is still not NULL, we have an excess of arguments, so print
+         * out a nice diagnostic message.
+         */
+        fprintf(stderr, "%s: ignoring extraneous non-option arguments: ",
+            getprogname());
+        for (;;) {
+            fputs(*argv, stderr);
+            if (*++argv)
+                fputc(' ', stderr);
+            else
+                break;
+        }
+        fputc('\n', stderr);
+    }
+    if (strcmp(opt.outputFile, "-") == 0) {
+        opt.overwrite = 1;
+        opt.thumb = THUMB_DISABLED;
+        opt.outputFile = getPathOfStdout();
+    }
+    if (opt.thumb != THUMB_DISABLED)
+        opt.thumbFile = optionsNameThumbnail(opt.outputFile);
 }
 
 static void showUsage(void)
 {
     fputs(/* Check that everything lines up after any changes. */
         "usage:  " PACKAGE " [-bcfhimopuvz] [-a X,Y,W,H] [-C NAME] [-D DISPLAY]\n"
-        "              [-d SEC] [-e CMD] [-F FILE] [-k OPT] [-l STYLE] [-M NUM]\n"
-        "              [-n OPTS] [-q NUM] [-S CMD] [-s OPTS] [-t NUM | GEOM] [FILE]\n",
+        "              [-d SEC] [-e CMD] [-k OPT] [-l STYLE] [-M NUM] [-n OPTS]\n"
+        "              [-q NUM] [-S CMD] [-s OPTS] [-t % | WxH] [[-F] FILE]\n",
         stdout);
     exit(0);
 }
@@ -536,7 +544,7 @@ static void showVersion(void)
     exit(0);
 }
 
-char *optionsNameThumbnail(const char *name)
+static char *optionsNameThumbnail(const char *name)
 {
     const size_t nameLength = strlen(name);
     const char thumbSuffix[] = "-thumb";
@@ -621,13 +629,6 @@ static void optionsParseThumbnail(char *optarg)
                 errmsg);
         }
     }
-}
-
-void optionsParseFileName(const char *optarg)
-{
-    checkMaxOutputFileName(optarg);
-    free(opt.outputFile);
-    opt.outputFile = estrdup(optarg);
 }
 
 void optionsParseNote(char *optarg)
--- a/configure.ac
+++ b/configure.ac
@@ -34,7 +34,8 @@ PKG_CHECK_MODULES([IMLIB2], [imlib2])
 AC_ARG_ENABLE([libbsd-feature-test],
     AS_HELP_STRING([--enable-libbsd-feature-test],
         ["Do not configure the program, return true if libbsd is needed"]))
-AC_CHECK_FUNCS([strlcpy strlcat err errx warn warnx],, [LIBBSD_NEEDED=yes])
+AC_CHECK_FUNCS([strlcpy strlcat err errx getprogname warn warnx],,
+    [LIBBSD_NEEDED=yes])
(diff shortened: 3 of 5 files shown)


N-R-K:
`-` is a valid filename, though. I'd keep `-F -` which outputs to filenamed `-` and only do the `-` special casing on non-option argument.

guijan:
For that you can output to __./-__

N-R-K:
> For that you can output to **./-**

Ah yes, forgot about that.

N-R-K:
Doesn't seem to work though, as of current `master` at least:

```console
[scrot master]~> ./src/scrot ./-
scrot: Saving to file ./- failed: No such file or directory
[scrot master]~> ./src/scrot -F ./-
scrot: Saving to file ./- failed: No such file or directory
```

Will open a bug about this later (unless I'm making some obvious mistake here).

guijan:
Every time we touch the code, a dozen bugs fall out...
This PR doesn't fix it either.

The error message comes from these lines:
https://github.com/resurrecting-open-source-projects/scrot/blob/e332c8b3982d53f6c5a26f0550ed80cb2796a807/src/scrot.c#L158-L160
Yet another error is that imlib_save_image_with_error_return() isn't documented to set errno on error, so `err()` should be `errx()`. The "No such file or directory" message is probably because errno is already set to that due to the `access()` calls before that code, I haven't checked to be certain.

It may well be that the error you found is inside imlib, too.

guijan:
Send it in a different PR, this one has a merge conflict with #217

N-R-K:
> Send it in a different PR, this one has a merge conflict with #217

Aright, opened #232 (though I found _another_ separate memory leak in the process, will fix soon).

guijan:
This still needs work:
```console
$ src/scrot -
scrot: Saving to file /dev/stdout_000 failed: Permission denied
```
I've only fixed the merge conflicts.

Adding stdout support to -F is not so simple because you can specify it multiple times, something has to be done to preserve the default of not overwriting in a sequence like this:
```sh
scrot -F/tmp/A.png -F- -F/tmp/B.png
```
There's probably more to it.

guijan:
Didn't expect it to be so easy, I've tested the latest push with the tests in the starting comment and also these:
```console
$ src/scrot -F- > /tmp/file.png
$ src/scrot - > /tmp/file.png
```
Both worked.

man/scrot.txt
```diff
@@ -3,8 +3,8 @@ NAME
 
 SYNOPSIS
   scrot [-bcfhimopuvz] [-a X,Y,W,H] [-C NAME] [-D DISPLAY] [-d SEC] [-e CMD]
-        [-F FILE] [-k OPT] [-l STYLE] [-M NUM] [-n OPTS] [-q NUM] [-S CMD] [-s OPTS]
-        [-t % | WxH] [FILE]
+        [-k OPT] [-l STYLE] [-M NUM] [-n OPTS] [-q NUM] [-S CMD] [-s OPTS]
+        [-t % | WxH] [FILE | -F FILE]
```
N-R-K, line 7: This looks really weird. I don't think I've ever seen non-option arg and option arg be mixed like this before.

man/scrot.txt
```diff
@@ -68,8 +67,11 @@ OPTIONS
   -u, --focused             Use the currently focused window.
   -v, --version             Output version information and exit.
   -z, --silent              Prevent beeping.
-  -                         Redirection to standard output. The output image
-                            format is PNG.
+  FILE | -F, --file FILE    Specify the output file. A "-" output file outputs
```
N-R-K, line 70: Same as above, I'd revert this and just put in the description that -F and FILE are mutually exclusive.

N-R-K (changes requested):

src/options.c
```diff
@@ -484,27 +481,38 @@ void optionsParse(int argc, char *argv[])
...
+        }
+        fputc('\n', stderr);
+        exit(EXIT_FAILURE);
+    }
+    if (strcmp(opt.outputFile, "-") == 0) {
+        opt.overwrite = 1;
+        opt.thumb = THUMB_DISABLED;
+        opt.outputFile = (/* const */ char *)getPathOfStdout();
+    }
+    opt.outputFile = estrdup(opt.outputFile);
```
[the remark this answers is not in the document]
guijan: I noticed that while working on it. There's no reason, actually. `opt.outputFile` gets passed around a little, and is only ever read but not written. It's not `free()`d either. 

I decided to keep things as they are because in the end, it's used by `imPrintf()`, and that function needs some fixing. It outputs to a 4KiB buffer on the stack and then duplicates it. While making this PR, I thought it would be best if `imPrintf()` was modified to assume its `str` parameter was alloc storage duration and grow it as needed, but I just realized a better design would be if the `strdup()` happened as late as possible which is inside `imPrintf()`.

man/scrot.txt
guijan: How about imitating st's manual here then? https://manpages.debian.org/bullseye/stterm/st.1.en.html

It has an `-e` option which specifies a command for the terminal to run on startup. It can also be specified as a non-option argument, so the manual's synopsis is `[[-e] command [arguments...]]`.

man/scrot.txt
guijan: Maybe it would be good to imitate st's manual here too? st's manual only has the equivalent of the `-F, --file FILE` bit, and the text beside it explains that a non-option argument works as if it was the option.

man/scrot.txt
N-R-K: > How about imitating st's manual here then?

ST manual is a bit different than this though. `-e` in ST isn't mutually exclusive with non-option argument.

But what exactly do you have in mind? I think I can give a better answer if I see a more concrete example.

N-R-K: I guess `[[-F] FILE]` would be ok.

man/scrot.txt
N-R-K: That seem fine. Would need to explain that `-` acts as stdout in the text also.

src/options.c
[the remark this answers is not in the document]
N-R-K: If there's no need to dup, then I'd drop the `estrdup` and make `outputFile` const.

guijan:
Requested changes done and commits squashed. 

Testing with this script:
```sh
src/scrot -F test test2; printf 'scrot returned: %s\n' $?
src/scrot -F test test2 test3; printf 'scrot returned: %s\n' $?
src/scrot test test2; printf 'scrot returned: %s\n' $?
src/scrot test test2 test3 test4; printf 'scrot returned: %s\n' $?
src/scrot "$(yes | tr -d '\n' | dd bs=251 count=1 2>/dev/null)".png; printf 'scrot returned: %s\n' $?
src/scrot "$(yes | tr -d '\n' | dd bs=256 count=1 2>/dev/null)".png; printf 'scrot returned: %s\n' $?
```

Console text dump:
```console
$ src/scrot -F test test2; printf 'scrot returned: %s\n' $?
scrot: both the --file option and the FILE parameter were used
scrot returned: 1
$ src/scrot -F test test2 test3; printf 'scrot returned: %s\n' $?
scrot: both the --file option and the FILE parameter were used
scrot returned: 1
$ src/scrot test test2; printf 'scrot returned: %s\n' $?
scrot: extraneous non-option arguments: test2
scrot returned: 0
$ src/scrot test test2 test3 test4; printf 'scrot returned: %s\n' $?
scrot: extraneous non-option arguments: test2 test3 test4
scrot returned: 0
$ src/scrot "$(yes | tr -d '\n' | dd bs=251 count=1 2>/dev/null)".png; printf 'scrot returned: %s\n' $?
scrot returned: 0
$ src/scrot "$(yes | tr -d '\n' | dd bs=256 count=1 2>/dev/null)".png; printf 'scrot returned: %s\n' $?
scrot: Saving to file yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy.png failed
scrot returned: 1
```
Everything looks fine from the testing but this PR sat for 2 weeks and I don't really remember what I was doing then, so there might still be something wrong with it.

man/scrot.txt
```diff
@@ -72,8 +71,12 @@ OPTIONS
   -w, --window              Window identifier to capture.
                             Must be a valid identifier (see xwininfo(1)).
   -z, --silent              Prevent beeping.
-  -                         Redirection to standard output. The output image
-                            format is PNG.
+  -F, --file FILE           Specify the output file. A "-" output file outputs
+                            a PNG image to stdout. The filename is expanded
```
N-R-K, line 75: Change 
> A "-" output file outputs a PNG image to stdout.

to
>If FILE is "-", scrot will output a PNG image to stdout.

N-R-K (changes requested):
Some small tweaks

src/options.c
```diff
@@ -497,35 +494,44 @@ void optionsParse(int argc, char *argv[])
...
-            if (redirectChar) {
-                optionsParseFileName(getPathOfStdout());
-                opt.overwrite = 1;
-                opt.thumb = THUMB_DISABLED;
-            }
-        } else
-            warnx("unrecognised option %s", argv[optind++]);
+    argv += optind;
+
+    if (*argv) {
```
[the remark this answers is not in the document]
guijan: I implemented with the semantics you suggest, but that has the weird semantics of an earlier argument (-F FILE) overriding the later non-option argument (FILE). Also, these weird semantics are more lines of code. 

I think it's better if I let the non-option argument override -F, is that fine?

guijan:
Will do testing when I wake up tomorrow, now I'm too sleepy, I already got the commit message wrong:
```plain
Improve the wording of the warning presented when the output file has
been specified through both -F and a non-option argument.
```
Should have been:
```plain
Improve the wording of the warning presented when extraneous non-option arguments are present.
```

src/options.c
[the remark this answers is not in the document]
N-R-K: The convention for cli flags is that the later argument overrides earlier one. Which can be handy for wrapper scripts like this:

```sh
prog --flag=custom-default "$@"
```

But when both -F and non-option is specific (which in scrot's context have the same meaning) is weird to deal with, indeed. This gets even more weird when you realize that GNU getopt permutes the argument to bring flags before. E.g

```console
$ prog some-non-opt -F some-opt
```

On glibc systems, the `-F some-opt` will be parsed first where on musl it will stop option parsing after it finds the first non-opt.

I'll have to think about this a bit, more. I'm at the end of my awake hours too, so don't want to put much thought into it at this state :)

src/options.c
```diff
@@ -386,6 +385,7 @@ void optionsParse(int argc, char *argv[])
     };
     int optch = 0;
     const char *errmsg;
+    int FFlagSet = 0;
```
N-R-K, line 388: Change it to `bool` and `false`

N-R-K (changes requested):
After thinking a bit more, I think the current state is fine. If someone puts both -F and non-opt, just let -F take priority and ignore the non-opt. If overwriting an earlier -F is desired, you can use another -F at the end.

src/options.c
guijan: We have a mixture of bool and int in the code, better decide which will be used and explain it in the CONTRIBUTING.md at some point. I don't like bool but I'd be indifferent here just to have one project that uses it and get some practice.
https://github.com/resurrecting-open-source-projects/scrot/blob/7ee199cd9729b8887ee5574ec899374226d18bda/src/options.c#L440-L454

guijan:
Tested with the shell script in
https://github.com/resurrecting-open-source-projects/scrot/pull/229#issuecomment-1494935840.

Console text dump:
```console
$ src/scrot -F test test2; printf 'scrot returned: %s\n' $?
scrot: ignoring extraneous non-option arguments: test2
scrot returned: 0
$ src/scrot -F test test2 test3; printf 'scrot returned: %s\n' $?
scrot: ignoring extraneous non-option arguments: test2 test3
scrot returned: 0
$ src/scrot test test2; printf 'scrot returned: %s\n' $?
scrot: ignoring extraneous non-option arguments: test2
scrot returned: 0
$ src/scrot test test2 test3 test4; printf 'scrot returned: %s\n' $?
scrot: ignoring extraneous non-option arguments: test2 test3 test4
scrot returned: 0
$ src/scrot "$(yes | tr -d '\n' | dd bs=251 count=1 2>/dev/null)".png; printf 'scrot returned: %s\n' $?
scrot returned: 0
$ src/scrot "$(yes | tr -d '\n' | dd bs=256 count=1 2>/dev/null)".png; printf 'scrot returned: %s\n' $?
scrot: Saving to file yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy.png failed
scrot returned: 1
```
Images created:
```console
$ git status
On branch same-dir
Your branch and 'origin/same-dir' have diverged,
and have 1 and 2 different commits each, respectively.
  (use "git pull" to merge the remote branch into yours)

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        test
        test_000
        test_001
        test_002
        yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy
yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy
yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy.png

nothing added to commit but untracked files present (use "git add" to track)
```


More testing for the thumbnail code with this script:
```sh
src/scrot -t 50x50
src/scrot -F test.png -t 50x50
```
```console
$ git status
On branch same-dir
Your branch and 'origin/same-dir' have diverged,
and have 1 and 2 different commits each, respectively.
  (use "git pull" to merge the remote branch into yours)

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        2023-04-04-160113_1920x1080_scrot.png
        2023-04-04-160113_50x50_scrot-thumb.png
        test-thumb.png
        test.png

nothing added to commit but untracked files present (use "git add" to track)
```

N-R-K (approved):

N-R-K:
Thanks.

N-R-K:
The CI [seems to be failing](https://github.com/resurrecting-open-source-projects/scrot/runs/12514848264) since this merge, but it looks unrelated to the PR. Some sort of alpine issue, it seems.

PR #231, opened 2023-03-17
remove -S flag

guijan:
Relevant imlib documentation:
https://docs.enlightenment.org/api/imlib2/html/
CTRL+F "Imlib 2 Dynamic Filters"

The -S flag makes imlib filters available to the user. However, it's impossible to do this, because imlib filters can reference variables passed to `imlib_apply_filter()` via varargs, and there's no way in C to pass the right amount of varargs. Look up "[] operator" in the imlib documentation.

Additionally, this option may be the source of a security vulnerability. A malicious imlib script could decrement the stack pointer by using the right amount of imlib script [] operators. Varargs are typically implemented by popping values off the stack. By changing what the stack points to, the values of variables or the return address could be changed, the stack pointer could even be decremented until it points all the way to the argument vector (which is in the stack on Linux at least, I don't know about other platforms) and that is also controlled by the program's caller.

This option is broken, impossible to fix, a possible security vulnerability, and obscure enough that it doesn't necessarily have users, so let's remove it.

### DIFF
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -473,6 +473,8 @@ end:
 // It assumes that the local variable 'scrot.c:Imlib_Image image' is in context
 static void applyFilterIfRequired(void)
 {
+    warnx("--script is deprecated. See: "
+        "https://github.com/resurrecting-open-source-projects/scrot/pull/231");
     if (opt.script)
         imlib_apply_filter(opt.script);
 }
--- a/man/scrot.txt
+++ b/man/scrot.txt
@@ -3,8 +3,8 @@ NAME
 
 SYNOPSIS
   scrot [-bcfhimopuvz] [-a X,Y,W,H] [-C NAME] [-D DISPLAY] [-d SEC] [-e CMD]
-        [-k OPT] [-l STYLE] [-M NUM] [-n OPTS] [-q NUM] [-S CMD] [-s OPTS]
-        [-t % | WxH] [-w NUM] [[-F] FILE]
+        [-k OPT] [-l STYLE] [-M NUM] [-n OPTS] [-q NUM] [-s OPTS] [-t % | WxH]
+        [-w NUM] [[-F] FILE]
 
 DESCRIPTION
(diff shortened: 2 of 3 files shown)


guijan:
The latest force-push fixes Merge conflicts. The commit message is also slightly changed for brevity.

N-R-K:
I _do_ support removing unnecessary cruft if there's no users for that feature.

However, since scrot is a very mature software, we need to be far more careful about it. Just removing something out of the blue is not a good idea, we don't _really_ know if there's users or not.

I'm thinking of something along the following:

* Keep the feature, but remove it from the docs
* Add a warning message if the flag is used. Something like, "this feature is subject to removal, please provide feedback if this is something that affects you at: \<link\>"
* Then wait 4~6 months after a release.

If we don't get any feedback during that time period, then we can be reasonably sure that there aren't many (or any) users of that feature and can remove it afterwards in the next release.

guijan:
I think it should be removed regardless of what users think because it's unfixable short of including imlib's parser for its scripting language in scrot. But sure, I'll change this to a warning.

src/scrot.c
```diff
@@ -465,6 +465,7 @@ void scrotGrabMousePointer(const Imlib_Image image, const int xOffset,
 // It assumes that the local variable 'scrot.c:Imlib_Image image' is in context
 static void applyFilterIfRequired(void)
 {
+    warnx("--script is deprecated. Read the man page for details");
```
N-R-K, line 468: The point of the warning is to see if anyone actually uses this option or not. And to warn them if they do. Doesn't make any sense to keep it around in the manpage if we're planning to remove it.

Just remove the manpage entry entirely and change the warning to:

```suggestion
    warnx("--script is deprecated. See: https://github.com/resurrecting-open-source-projects/scrot/pull/231");
```

N-R-K (changes requested):

N-R-K (approved):
I'll wait a bit before merging. Input from @daltomi would be appreciated.

PR #232, opened 2023-03-18
Simplify getPathOfStdout()

N-R-K:
let the caller duplicate it if needed, allows for passing the return
value straight to optionsParseFileName()

also since STDOUT_FILENO is defined to be 1 there was no need to call
snprintf().

- - -

Includes the mem leak fix from #229

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -336,26 +336,15 @@ static bool accessFileOk(const char *const pathName)
     return (0 == access(pathName, W_OK));
 }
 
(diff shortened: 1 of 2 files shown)


guijan (approved):
I haven't actually tested it, but a careful reading says the changes are good.

N-R-K:
I'll give this a day or two before merging, giving anyone else who might be interested a chance to comment/review.

PR #234, opened 2023-03-18
scrotGetWindowName(): change to internal linkage

guijan:
Also moves the definition of the function so the functions in the source file are defined closer to the order in which they're called.


src/scrot.c
```diff
@@ -717,6 +692,33 @@ static char *imPrintf(char *str, struct tm *tm, char *filenameIM,
...
+
+    if (status != 0) {
+        windowName = estrdup(clsHint.res_class);
+        XFree(clsHint.res_name);
+        XFree(clsHint.res_class);
+    }
+    return windowName;
+}
+
+
```
N-R-K, line 721: Probably an unintended double space from copy-pasting, but otherwise looks OK.

N-R-K (approved):

PR #238, opened 2023-03-18
add --delay-select

N-R-K:
multiple users have expressed wanting to have the selection occur after the delay. this adds a flag which provides that behavior.

Closes: https://github.com/resurrecting-open-source-projects/scrot/issues/221

### DIFF
--- a/man/scrot.txt
+++ b/man/scrot.txt
@@ -30,8 +30,8 @@ OPTIONS
   -c, --count               Display a countdown when used with -d.
   -D, --display DISPLAY     DISPLAY is the display to use; see X(7).
   -d, --delay SEC           Wait SEC seconds before taking a shot.
-      --delay-select        If delay is active, then delay the selection as
-                            well.
+      --delay-select        If both delay and selection is active, then
+                            perform the delay first.
   -e, --exec CMD            Execute CMD on the saved image.
   -F, --file                File name. See SPECIAL STRINGS.
   -f, --freeze              Freeze the screen when -s is used.
--- a/man/scrot.txt
+++ b/man/scrot.txt
@@ -30,6 +30,8 @@ OPTIONS
   -c, --count               Display a countdown when used with -d.
   -D, --display DISPLAY     DISPLAY is the display to use; see X(7).
   -d, --delay SEC           Wait SEC seconds before taking a shot.
+      --delay-select        If delay is active, then delay the selection as
+                            well.
   -e, --exec CMD            Execute CMD on the saved image.
   -F, --file                File name. See SPECIAL STRINGS.
   -f, --freeze              Freeze the screen when -s is used.
--- a/src/options.c
+++ b/src/options.c
@@ -361,6 +361,11 @@ static char *getPathOfStdout(void)
 void optionsParse(int argc, char *argv[])
 {
     static char stropts[] = "a:ofipbcd:e:hmq:s::t:uvzn:l:D:k::C:S:F:M:";
+    enum { /* constants for long-opt only */
+        /* ensure these cannot be represented as a single byte so that they
+         * don't collude with short-opts */
+        OPT_DELAY_SELECT = UCHAR_MAX + 1,
+    };
 
     static struct option lopts[] = {
         /* actions */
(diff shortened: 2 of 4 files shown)


guijan:
Maybe that could be a suboption to `-s`.
```console
scrot -d 5 -s after-delay
```
That would delay for 5 seconds, and show the selection interface after the delay.

N-R-K:
> Maybe that could be a suboption to `-s`.

Probably. But looking at `SELECTION MODE` section of the manpage, the `-s` option already feels too over-complicated for my taste.

Although I'm not entirely against the idea. I'd like to hear some more user input on this if possible.

cc: @kucharskim

N-R-K:
> Maybe that could be a suboption to `-s`.

Tried it out, but it gets pretty awkward to even document, especially since it wouldn't be mutually exclusive with the existing sub-options.

Having an obvious long-opt seems much better interface wise.

kucharskim:
I don't know, would need to think, and all this seems already over complicated in my mind, however after being pointed to FAQ I get why. I would need to re-read the man page, the FAQ, but I don't really have time. Long option is too much typing for me, so I still probably will stick to `sleep 5; scrot -s` as it is very explicit what it does.

kucharskim:
Interestingly pull request of `--delay-select` moves when `-d <SECONDS>` is applied, before or after the select. I was thinking to be flexible and have both options of the behavior, so `-d <NUM>` stays as is, and (short option is just an example, I know it's still in use) is via `-S <SECONDS>` so you can apply both at the same time, `scrot -S 5 -d 10 ...`. I didn't think through this too much, it was more an reflex thought, without simulating everything in my head. It feels just simpler to me. After I've looked at the code of `--delay-select` how it works, and how many letters to type, and that still `-d <NUM>` is needed, I cannot see myself using it :shrug: Sorry, I am just being honest.

kucharskim:
I probably should be more clear when I typed the example, I wrote in the issue:

      -E, --delay-select

but what I SHOULD write was:

      -E <SECONDS>, --delay-select <SECONDS>

Sorry, by not being explicit end precise with my comment on the issue, you probably ended with this pull request :/

N-R-K:
> so you can apply both at the same time, `scrot -S 5 -d 10`

Hmm, that makes sense. And I'll admit that `--delay-select` as it currently is, is kinda awkward. It only makes sense when *both* delay and select are already active - I thought about making `--delay-select` automatically enable select as well but unfortunately `--select` is not a binary flag, it takes optional arguments.

Here's another idea, how about making the `-d` argument accept a `s` prefix. `-d s8` would mean delay selection by 8 seconds, and `-d 8` would work as usual.

I'm not sure if there's any use-case that needs to combine both or not, so I'd avoid unnecessarily adding features no-one would use. But if needed then this idea can be extended with a comma separator too, something like `-d s8,4` would delay selection by 8 sec and then 4 more seconds of delay before capture.

kucharskim:
>      -d s8

I was thinking similar, but as negative number, is it before or after, so:

      -d -8

However I am not sure. If I would see `s8` in man page my first tought would be, is it a typical typo / swap of characters error and should it be `8s` - eight seconds? Then it needs to be clearly explained in the man page, that `s` is there, before on purpose.

N-R-K:
```
  -d, --delay [s]SEC        Wait SEC seconds before taking a shot.
                            When given the `s` prefix, e.g `-d s8`, the delay
                            will be applied before selection.
```

This is what I have in mind, the square brackets are conventional way to indicate something is optional. But we can probably use a different letter other than `s` if that's confusing. Perhaps `b` for *before* selection.

kucharskim:
Sure, whatever works. I personally would go for `-` and `+` when `+` is default when skipped, but I think `b` for _before_ also make sense. I think `b` is less confusing than `s`.

N-R-K:
One can argue that a "negative delay" is confusing too. `b` seems like a good choice. I'll go with that. Opened https://github.com/resurrecting-open-source-projects/scrot/pull/254.

Let's continue there if there's any other feedback you want to provide (and btw, thanks for the feedback, it's been helpful).

PR #242, opened 2023-03-19
fix multiple bugs in scrot's -C option

guijan:
I've carefully read the Xlib manuals and some more, but this PR should be reviewed by someone more familiar with X programming.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -73,7 +73,6 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
     STR_LEN_MAX_FILENAME(input, (fileName))
 
 enum {
-    MAX_LEN_WINDOW_CLASS_NAME = 80, //characters
     MAX_FILENAME = 256, // characters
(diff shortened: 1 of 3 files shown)


src/options.c
```diff
@@ -474,7 +465,7 @@ void optionsParse(int argc, char *argv[])
             optionsParseStack(optarg);
             break;
         case 'C':
-            optionsParseWindowClassName(optarg);
+            opt.windowClassName = optarg;
```
N-R-K, line 468: Could you make `opt.windowClassName` const in options.h ? Assuming there won't be any issues.

N-R-K:
Unfortunately I cannot test this since the `-k` options seems to require a compositor (and I don't want to go though the hassle of setting one up).

Can you confirm that you did some before/after testing (just in case) @guijan ? If yes, then I'll go ahead and merge this since the code looks OK to me.

guijan:
Yeah, I've tested it. Here's a screenshot I just took.
2023-03-25-073826_956x2144_scrot

N-R-K:
Thanks.

Issue #226, opened 2023-03-17
Screenshot and thumbnail not created in the same directory


guijan:
Consider the following invocation:
```sh
scrot --thumb 10 /tmp/test.png
```

Scrot's execution would then be like this:

First, we'd we'd end up on this line:
https://github.com/resurrecting-open-source-projects/scrot/blob/f1b195fb8dc56f7dbc7f349bf3d0b2ba41e1af62/src/scrot.c#L158
The string `filenameIM` points to would be `"/tmp/test.png"`

A little further down, we'd end up on this line of code:
https://github.com/resurrecting-open-source-projects/scrot/blob/f1b195fb8dc56f7dbc7f349bf3d0b2ba41e1af62/src/scrot.c#L190
The string `filenameThumb` points to would be `"/tmp/test-thumb.png"`

The bug here is that between those 2 lines of code, __/tmp/__ could be changed to point to a different directory.

The solution should be to `chdir()` to the output file's directory, then handle the output file and thumbnail file as basenames only.
This also has the nice side effect of allowing the creation output files with longer filenames: as the code currently stands, With a PATH_MAX of 4096 as is typical these days, and the directory portion of the filename alone is 4000 bytes, we may be unable to create a file with a basename over 95 bytes (1 byte reserved for the '\0') even though X/Open guarantees NAME_MAX is at least 255.

N-R-K:
> The solution should be to `chdir()` to the output file's directory, then handle the output file and thumbnail file as basenames only.

Or we can also just open the directory and then use `openat` to create/open the files. See commit https://github.com/resurrecting-open-source-projects/scrot/commit/c6f91383c798bd5184d10bf81930959dcc751873 in the `v2` branch for context.

However, when trying to do this, I realized that the directory of the file and the thumbnail need not be the same. Example:

```console
$ scrot -t16 '/tmp/$w/shot.png'
```

Will create the thumbnail and the screenshot in different directory since the image width is used in the directory - which will likely be different for the screenshot vs the thumbnail.

Issue #241, opened 2023-03-18
Saving to `./-` does not work


N-R-K:
```console
[scrot master]~> ./src/scrot ./-
scrot: Saving to file ./- failed: No such file or directory
[scrot master]~> ./src/scrot -F ./-
scrot: Saving to file ./- failed: No such file or directory
```

Using `-F -` however works fine.

_Originally posted at: https://github.com/resurrecting-open-source-projects/scrot/pull/229#issuecomment-1474518175_

N-R-K:
Looks like this will be more tricky than imagined. If someone uses `a.com` as the output filename, then we won't set any format and imlib2 will fail since it doesn't have any `.com` savers.

I was thinking it might be good idea to retry with PNG set if the from imlib2 is due to not having a saver for that format. However, then in cases like `file.webp` the user most likely is going to expect a webp output. But if imlib2 doesn't have webp saver, and we end up silently saving it as PNG, then that's unexpected.

I don't think relying on extension to get the output file format was a good idea. We should probably just add a `--format` flag so users can more explicitly set the output format.

Disonantemus:
Using: `scrot --format webp` resulting file was `.png`, expected was `.webp`, checking with medianinfo & ffmpeg says **webp**:
~~~
> mediainfo 2023-04-28-113049_1920x1080_scrot.png
General
Complete name                            : 2023-04-28-113049_1920x1080_scrot.png
Format                                   : WebP
File size                                : 88.2 KiB

Image
Format                                   : WebP
~~~ 

~~~
> ffprobe -hide_banner 2023-04-28-113049_1920x1080_scrot.png
Input #0, webp_pipe, from '2023-04-28-113049_1920x1080_scrot.png':
  Duration: N/A, bitrate: N/A
  Stream #0:0: Video: webp, yuv420p(tv, bt470bg/unknown/unknown), 1920x1080, 25 fps, 25 tbr, 25 tbn
~~~

It's safe to just rename to webp?

Issue #228, opened 2023-03-17
`scrot /tmp/` creates "/tmp/_000"


guijan:
The code checks if __/tmp/__ already exists, and because it does, it appends ___000__ and creates that instead. We can't create the file because a directory by that name already exists, so we create a file inside it? This is a funny quirk rather than something intentional. 

The code should probably error in this case.

But what should we do about `cd / && scrot tmp`? Nothing, possibly?

N-R-K:
> But what should we do about `cd / && scrot tmp`? Nothing, possibly?

Attempting `tmp_000` is *probably* not a bad idea in that case. I can imagine someone doing something like `scrot test` trying to create a file `test` but it already might exist as a directory. In which case `test_000` isn't *too* unexpected of an outcome.

OOTH if someone invokes scrot with a trailing slash (e.g `/tmp/`), then we can:

A. Error out early (similar to #361)
B. Strip all trailing slashes from the filename so it behaves as if `/tmp` was specified in the cli.

I'm in favor of (A) because it seems less surprising than having your explicit trailing slash stripped silently.

N-R-K:
I guess there's also option C

C. Interpret a trailing slash as a output dir. E.g `scort /tmp/` would create a file inside of `/tmp` with the default output format.

Personally, I don't like this. It overloads the output filename with yet another functionality (currently we use output filename's extension to determine format also).

- - -

(This also made me realize that scrot's extension detection is still broken: `scrot ~/.config/testfile` will end up trying to use `.config/testfile` as `opt.format`).

PR #248, opened 2023-03-20
clean up optionsNameThumbnail()

guijan:
Sorry but this function has been bugging me ever since #220. I knew it could be better but I failed back then.

I've tested it with these commands:
```sh
src/scrot --thumb 50 test.png
src/scrot --thumb 50 test
src/scrot --thumb 50 te.st.png
```
As you can see, all the right files were created:
```console
$ git status
On branch briefer-namethumb
Your branch is up to date with 'origin/briefer-namethumb'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        te.st-thumb.png
        te.st.png
        test
        test-thumb
        test-thumb.png
        test.png

nothing added to commit but untracked files present (use "git add" to track)
```

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -525,25 +525,19 @@ static void showVersion(void)
 
 char *optionsNameThumbnail(const char *name)
 {
+    const size_t nameLength = strlen(name);
     const char thumbSuffix[] = "-thumb";
-    const size_t newNameLength = strlen(name) + sizeof(thumbSuffix);
-    char *newName = malloc(newNameLength);
-
-    if (!newName)
-        err(EXIT_FAILURE, "Unable to allocate thumbnail");
-
+    Stream ret = {0};
     const char *const extension = strrchr(name, '.');
+    const size_t baseNameLength = extension ? extension-name : nameLength;
 
-    if (extension) {
-        /* We add one so length includes '\0'*/
-        const ptrdiff_t nameLength = (extension - name) + 1;
-        strlcpy(newName, name, nameLength);
-        strlcat(newName, thumbSuffix, newNameLength);
-        strlcat(newName, extension, newNameLength);
-    } else
-        snprintf(newName, newNameLength, "%s%s", name, thumbSuffix);
+    streamMem(&ret, name, baseNameLength);
+    streamMem(&ret, thumbSuffix, sizeof(thumbSuffix)-1);
+    if (extension)
+        streamMem(&ret, extension, nameLength - baseNameLength);
+    streamChar(&ret, '\0');
 
-    return newName;
+    return ret.buf;
 }
 
 void optionsParseAutoselect(char *optarg)


guijan:
Force-push because I had an extra `;` of all things.

N-R-K:
I don't really like the existing function either, but the improvement here is very marginal. It's still directly messing with pointer arithmetic which are error prone.

I honestly had no clue that a screenshot tool would have so much string manipulation going on! If I were to write scrot from scratch then I definitely would've opted for using sized-strings, with "higher level" string manipulation routines.

I also wouldn't have created a massive web of allocated pointers, each of which needs individual managing, which is error prone (evident by the amount of leaks we've discovered lately). Instead the strings would go into [an arena](https://www.rfleury.com/p/untangling-lifetimes-the-arena-allocator) so that they can be managed [*in groups*](https://www.youtube.com/watch?v=f4ioc8-lDc0&t=4482s).

But in any-case, we're not writing scrot from scratch so we're far past that point :)

`</rant>`

- - -

But if this still bothers you, then you can use the stream functions from #250, they're guarded by assertions so mistakes are less likely to go unnoticed. Something like this (pseudo-code):

```c
Stream ret = {0};
streamReserve(newNameLen);

streamMem(name, ..);
streamMem(thumbSuffix, sizeof(thumbSuffix) - 1);
if (ext) streamMem(ext, extLen);
streamChar('\0');
return ret.buf;
```

Also, I'd probably use `memrchr` for the extension, it's portable enough and [is being considered](https://www.austingroupbugs.net/view.php?id=1633) for next POSIX.

guijan:
This version has a slightly better algorithm (less `strlcat()`'ing), generates smaller assembly on x64 with Clang, and reads like a clearer C version of an English explanation of what the function is supposed to do:
- Copy the string up to the extension (if any) exclusive
- Append "-thumb"
- Append the extension if we had one

It does, however, do a lot of working around the poor design of C's _string.h_. If you want, I can change it to this version which doesn't:
```c
char *optionsNameThumbnail(const char *name)
{
    char *ext = strrchr(name, '.');
    int basenameLen = ext ? ext-name : INT_MAX;
    char *ret;
    if (asprintf(&ret, "%.*s-thumb%s", basenameLen, name, ext ? ext : "") == -1)
        err(1, "asprintf");
    return ret;
}
```
Although I'm not sure how widely adopted `asprintf()` is. The "Stream" functions might be a good idea too, we can wait for #250. 

I considered `memrchr()`, but OS X doesn't have it, that would introduce a dependency on my libobsd library in the OS X version which @daltomi was against in the past.

Memory management in scrot isn't complicated, there's just a lot of bad code that needs fixing, but all we need to do is look for the last use of a variable and `free()` it. We don't have a complicated task we don't delegate to libraries to do, just unnecessarily complicated code. Hand-holding wouldn't make `malloc()` and `free()` simpler.

N-R-K:
> I considered `memrchr()`, but OS X doesn't have it,

Unfortunate, most other BSDs have it. Drop that idea then.

N-R-K:
One more thing I realized (just mentioning it so I don't forget about it later, doesn't need to be fixed in this PR) is that this function probably won't work reliably for cases like `./-` (#241) or basically any other pathname that has `.` or `..` it in.

guijan:
That's similar to #228.

I'd like to make the code `chdir()` to the output directory (acquired using `dirname()`) eventually which would also fix #226. Then it could error if the `basename()` of the output file is `"."`.

This would open the path for some mitigations that restrict the filesystem view like chroot, unveil, and landlock, but I'm not sure of how feasible those are because the output directory could contain `imPrintf()`'s special strings.

guijan:
Should be ready for review.

src/options.c
```diff
@@ -525,25 +525,19 @@ static void showVersion(void)
...
-        const ptrdiff_t nameLength = (extension - name) + 1;
-        strlcpy(newName, name, nameLength);
-        strlcat(newName, thumbSuffix, newNameLength);
-        strlcat(newName, extension, newNameLength);
-    } else
-        snprintf(newName, newNameLength, "%s%s", name, thumbSuffix);
+    streamMem(&ret, name, baseNameLength);
+    streamMem(&ret, thumbSuffix, sizeof(thumbSuffix)-1);
+    if (extension)
+	    streamMem(&ret, extension, nameLength - baseNameLength);
```
N-R-K, line 537: The indentation looks messed up here.

src/options.c
```diff
@@ -525,25 +525,19 @@ static void showVersion(void)
...
-        /* We add one so length includes '\0'*/
-        const ptrdiff_t nameLength = (extension - name) + 1;
-        strlcpy(newName, name, nameLength);
-        strlcat(newName, thumbSuffix, newNameLength);
-        strlcat(newName, extension, newNameLength);
-    } else
-        snprintf(newName, newNameLength, "%s%s", name, thumbSuffix);
+    streamMem(&ret, name, baseNameLength);
+    streamMem(&ret, thumbSuffix, sizeof(thumbSuffix)-1);
+    if (extension)
```
N-R-K, line 536: >don't pass NULL to streamMem
>
>This is actually perfectly valid but it looks weird.

The code is fine after this commit, but wanted to clarify that passing NULL to any of the string.h functions is UB _even if_ `n` is zero [^0] (and compilers like gcc/clang can aggressively optimize code around such calls with the assumptions that the pointer are non-null just because it was an argument in a memcpy call).

[^0]: https://port70.net/~nsz/c/c99/n1256.html#7.21.1p2

> Where an argument declared as size_t n specifies the length of the array for a function, n can have the value zero on a call to that function. Unless explicitly stated otherwise in the description of a particular function in this subclause, pointer arguments on such a call shall still have valid values

N-R-K (commented):
Didn't test (yet) but looks OK.

src/options.c
guijan: Quick fixes are the worst. That's a tab because I use tabs and have to set the editor to spaces every time. Fixed.

N-R-K:
Thanks.

Issue #223, opened 2023-02-26
Unnecessary string truncation in src/scrot.c


guijan:
This truncates:
https://github.com/resurrecting-open-source-projects/scrot/blob/4c59bd454f51401a22820d3a38973475da93cda4/src/scrot.c#L613-L614

sizeof(buf) is 20:
https://github.com/resurrecting-open-source-projects/scrot/blob/4c59bd454f51401a22820d3a38973475da93cda4/src/scrot.c#L596

[POSIX](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/limits.h.html) says:
```plain
{HOST_NAME_MAX}
    Maximum length of a host name (not including the terminating null) as returned from the [gethostname()](https://pubs.opengroup.org/onlinepubs/009695399/functions/gethostname.html) function.
    Minimum Acceptable Value: {_POSIX_HOST_NAME_MAX}
```
Further down:
```plain
{_POSIX_HOST_NAME_MAX}
    Maximum length of a host name (not including the terminating null) as returned from the [gethostname()](https://pubs.opengroup.org/onlinepubs/009695399/functions/gethostname.html) function.
    Value: 255
```

The fix here is to make buf's size HOST_NAME_MAX+1, but the real problem is the function's design. It should make gethostname() directly output to ret and it should dynamically grow ret as needed.

N-R-K:
Github doesn't understand "partially fixes" and keeps closing this, pretty annoying. :/

But in any case, `strftime` is still happening on a fixed length buffer. So I don't think this is fully fixed.

N-R-K:
I think the only remaining issue now is the `strftime` buffer. This would (almost) fix it:

```diff
     Stream ret = {0};
     long hostNameMax = 0;
-    char strf[4096];
+    char *strf = NULL;
+    size_t strfSize = 64;
     char *tmp;
     struct stat st;
 
-    if (strftime(strf, 4095, str, tm) == 0)
-        errx(EXIT_FAILURE, "strftime returned 0");
+    do {
+        strfSize *= 2;
+        strf = erealloc(strf, strfSize);
+    } while (strftime(strf, strfSize, str, tm) == 0); /* FIXME: infinite loop */
 
     imlib_context_set_image(im);
```

But the problem is strftime can validly return 0 and there's no way to distinguish between a valid 0 return vs a 0 return due to small buffer.

N-R-K:
From the `strftime` manpage:


[...]

PR #250, opened 2023-03-24
imPrintf: use dynamic buffer and concatenate efficiently

N-R-K:
The imPrintf function currently is horribly inefficient due to `strlcat` having to recompute `strlen` on _every single call_ [^0].

The requirement of nul-terminator also makes for very awkward code like these:

https://github.com/resurrecting-open-source-projects/scrot/blob/5388fffd7be700834ee888ed129b2325c8ab1ef2/src/scrot.c#L684-L685

Instead just remember where to append, and grow the buffer dynamically if needed (which would enable properly fixing https://github.com/resurrecting-open-source-projects/scrot/issues/223). 

[^0]:  If the standard string copying functions [weren't idiotic](https://www.symas.com/post/the-sad-state-of-c-strings) then strcat (and friends) wouldn't even need to exist.

### DIFF
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -583,12 +583,10 @@ static char *imPrintf(char *str, struct tm *tm, char *filenameIM,
 {
     char *c;
     char buf[20];
-    char ret[4096];
+    Stream ret = {0};
     char strf[4096];
     char *tmp;
     struct stat st;
-
-    ret[0] = '\0';
 
     if (strftime(strf, 4095, str, tm) == 0)
         errx(EXIT_FAILURE, "strftime returned 0");
@@ -600,32 +598,32 @@ static char *imPrintf(char *str, struct tm *tm, char *filenameIM,
             switch (*c) {
             case 'a':
                 gethostname(buf, sizeof(buf));
-                strlcat(ret, buf, sizeof(ret));
+                streamStr(&ret, buf);
                 break;
             case 'f':
                 if (filenameIM)
-                    strlcat(ret, filenameIM, sizeof(ret));
+                    streamStr(&ret, filenameIM);
                 break;
             case 'm': /* t was already taken, so m as in mini */
                 if (filenameThumb)
-                    strlcat(ret, filenameThumb, sizeof(ret));
+                    streamStr(&ret, filenameThumb);
                 break;
             case 'n':
                 if (filenameIM) {
                     tmp = strrchr(filenameIM, '/');
                     if (tmp)
-                        strlcat(ret, tmp + 1, sizeof(ret));
+                        streamStr(&ret, tmp + 1);
                     else
-                        strlcat(ret, filenameIM, sizeof(ret));
+                        streamStr(&ret, filenameIM);
                 }
                 break;
             case 'w':
                 snprintf(buf, sizeof(buf), "%d", imlib_image_get_width());
-                strlcat(ret, buf, sizeof(ret));
+                streamStr(&ret, buf);
                 break;
             case 'h':
                 snprintf(buf, sizeof(buf), "%d", imlib_image_get_height());
-                strlcat(ret, buf, sizeof(ret));
+                streamStr(&ret, buf);
                 break;
             case 's':
                 if (filenameIM) {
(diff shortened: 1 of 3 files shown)


N-R-K:
> 2 buffers just makes us have to keep track of and grow another buffer.

The only other buffer we need to (potentially) dynamically allocate is the one needed for host name. I can add a `streamReserve` so that we can `gethostname` directly into the stream (I was going to add `streamReserve()` anyways as a minor optimization to `streamMem`).

N-R-K:
I've pushed the change (lightly tested). Take a look and let me know how you feel about this.

src/scrot.c
```diff
@@ -606,33 +637,44 @@ static char *imPrintf(char *str, struct tm *tm, char *filenameIM,
...
+                    /* freebsd and macos doesn't have HOST_NAME_MAX defined.
+                     * instead sysconf is recommended by freebsd */
+                    if ((hostNameMax = sysconf(_SC_HOST_NAME_MAX)) < 0)
+                        errx(EXIT_FAILURE, "couldn't grab _SC_HOST_NAME_MAX");
+                    ++hostNameMax; /* for the nul terminator */
+                }
+
+                streamReserve(&ret, hostNameMax);
+                char *target = ret.buf + ret.off;
+                gethostname(target, hostNameMax);
```
guijan, line 650: One thing I noticed while writing my example was that maybe `{HOST_NAME_MAX}` could change between the `sysconf()` call and the `gethostname()` call. POSIX allows the value to change, but I don't know under what circumstances it may change or if it changes at all on a real OS. Maybe we should `memchr()` for the terminator and add our own if there's none to prevent a possible buffer overread. Let that edge case truncate instead of adding the fancy code to try again. Or ignore it and blame POSIX.

guijan:
It looks good, I didn't realize only `gethostname()` was troublesome, I hadn't looked very deeply into the function until now.

src/scrot.c
```diff
@@ -585,18 +590,44 @@ static Bool scrotXEventVisibility(Display *dpy, XEvent *ev, XPointer arg)
     return (ev->xvisibility.window == *win);
 }
 
+static void streamReserve(Stream *buf, size_t size)
+{
+    size_t avail = buf->cap - buf->off;
+    if (avail < size) {
+        buf->cap += MAX(size, 1024); /* be a bit greedy when allocating */
```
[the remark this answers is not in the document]
N-R-K: > Real paths created by users will be much smaller than 1KiB.

True. However, allocating too little will end up making unnecessary realloc calls. 128 should be a better lower-bound.

N-R-K:
Added some additional assertions. The patch is looking good, should be ready for review.

src/util.c
```diff
@@ -45,3 +46,40 @@ void *ecalloc(size_t nmemb, size_t size)
...
+void streamMem(Stream *buf, const char *str, size_t n)
+{
+    streamReserve(buf, n);
+    for (size_t i = 0; i < n; ++i)
+        buf->buf[buf->off++] = str[i];
+}
+
+void streamStr(Stream *buf, const char *str)
+{
+    streamMem(buf, str, strlen(str));
```
guijan, line 84: This should use `strlcpy()` and reserve more memory and try again if it truncated.

src/util.c
```diff
@@ -45,3 +46,40 @@ void *ecalloc(size_t nmemb, size_t size)
...
+{
+    void *p = realloc(ptr, size);
+    if (!p)
+        err(EXIT_FAILURE, "realloc");
+    return p;
+}
+
+void streamReserve(Stream *buf, size_t size)
+{
+    assert(buf->off <= buf->cap);
```
guijan, line 60: I don't like `assert()`. Either the check is worth putting in the release build or it's not worth having at all.

N-R-K: > Either the check is worth putting in the release build or it's not worth having at all.

Different tools for different purposes. Things like address sanitizer or even the `-g` flag of compiler are examples of tools that are meant to be used for debugging purposes. It doesn't make sense to say, "I won't use `-g` or sanitizers because it's not enabled in release build."

Assertions are no different in this regard, it's a debugging tool for debugging purposes - to catch a bug as early as possible instead of marching forward as if everything is OK.

- - -

On a sidenote, this also makes me wonder if we actually ever define `NDEBUG` or not. Probably makes sense to have it defined in the release tarballs (assuming it's not done already).

src/util.c
N-R-K: The logic that strlcpy propogates `copy => realloc => restart copy` is overcomplicated, unnatural and inefficient. "Reserve enough space and then copy over" is a significantly simpler and straightforward way to do things.

guijan: strlcpy() will iterate once if there's enough memory, the current code always iterates twice. This at least has the advantage of being short, so either way is fine I guess.

src/util.c
```diff
@@ -45,3 +46,40 @@ void *ecalloc(size_t nmemb, size_t size)
...
+    streamReserve(buf, 1);
+    buf->buf[buf->off++] = ch;
+}
+
+void streamMem(Stream *buf, const char *str, size_t n)
+{
+    streamReserve(buf, n);
+    for (size_t i = 0; i < n; ++i)
+        buf->buf[buf->off++] = str[i];
+}
```
guijan, line 80: I just noticed, we have `memcpy()` to copy memory. It's even likely this code gets optimized into a `memcpy()` call. And the 2nd parameter's type should be `void *` to show we're dealing with any memory.

src/util.c
N-R-K: > the current code always iterates twice

Unless the data is bigger than the L1d cache (unlikely), memory accesses are cheap. But the main difference between copying a nul-string vs copying a fixed len buffer is that the string copy's loop condition is typically some form of `s[i] != '\0'` which means the cpu needs to constantly keep checking for this condition and cannot (*observably*) execute the next loop iteration. But the copy len being known in the buffer case allows the processor to execute the next iteration before the first one finishes (due to optimizations such as [Out of Order execution](https://en.wikipedia.org/wiki/Out-of-order_execution)).

This is of course micro-optimization that's not relavant for our purposes (*main* efficiency gain in this patch is removal of strlcat _on each byte_).

src/util.c
N-R-K: Did it this way to keep the copy and `off` in sync in the same line. But I don't have strong preferences here, can change it to `memcpy`.

N-R-K: Fixed in new push.

PR #251, opened 2023-03-24
imPrintf: avoid hostname truncation

N-R-K:
this dynamically allocates a buffer for hostname if the local `buf` isn't big enough.

also change the `buf` size from 20 to 256 (which is HOST_NAME_MAX + 1) for most systems.


N-R-K:
Marking this as draft as the fix in #250 seems like a better solution.

N-R-K:
Closed due to #250

Issue #252, opened 2023-03-26
imPrintf goes out of bound


N-R-K:
If the last character of a filename is `$` or `\` then the `c` pointer can end up advancing past the end of the string.

https://github.com/resurrecting-open-source-projects/scrot/blob/7932f280d798cdb1d29174963e17cd41f6ce6691/src/scrot.c#L670-L672

Assuming `$` was the last character, this will push `\0` into `ret` and then `c` will be incremented past the end of the string on the next loop iteration.

guijan:
I haven't looked deeply into the code yet but a pointer to one past the end of an object isn't UB.

PR #253, opened 2023-03-26
imPrintf: avoid going out of bound

N-R-K:
when the last char is `$` or `\`, the `c` pointer would end up advancing past the string length.

additionally use the return value of strftime for determining the string length instead of checking for '\0'.

Fixes: https://github.com/resurrecting-open-source-projects/scrot/issues/252

### DIFF
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -590,7 +590,6 @@ static Bool scrotXEventVisibility(Display *dpy, XEvent *ev, XPointer arg)
 static char *imPrintf(char *str, struct tm *tm, char *filenameIM,
     char *filenameThumb, Imlib_Image im)
 {
-    char *c;
     char buf[20];
     Stream ret = {0};
     long hostNameMax = 0;
@@ -598,12 +597,13 @@ static char *imPrintf(char *str, struct tm *tm, char *filenameIM,
     char *tmp;
     struct stat st;
 
-    if (strftime(strf, 4095, str, tm) == 0)
+    const size_t strfLen = strftime(strf, sizeof(strf), str, tm);
+    if (strfLen == 0)
         errx(EXIT_FAILURE, "strftime returned 0");
 
     imlib_context_set_image(im);
-    for (c = strf; *c != '\0'; c++) {
-        if (*c == '$') {
+    for (const char *c = strf, *end = strf + strfLen; c < end; ++c) {
+        if (*c == '$' && (c + 1) < end) {
             c++;
             switch (*c) {
             case 'a':
@@ -680,7 +680,7 @@ static char *imPrintf(char *str, struct tm *tm, char *filenameIM,
                 streamChar(&ret, *c);
                 break;
             }
-        } else if (*c == '\\') {
+        } else if (*c == '\\' && (c + 1) < end) {
             c++;
             switch (*c) {
             case 'n':


src/scrot.c
```diff
@@ -581,22 +581,24 @@ static Bool scrotXEventVisibility(Display *dpy, XEvent *ev, XPointer arg)
...
         errx(EXIT_FAILURE, "strftime returned 0");
 
     imlib_context_set_image(im);
-    for (c = strf; *c != '\0'; c++) {
-        if (*c == '$') {
-            c++;
-            switch (*c) {
+    for (size_t i = 0; i < strfLen; ++i) {
+        char ch = strf[i];
+        if (ch == '$' && (i + 1) < strfLen) {
```
N-R-K, line 599: Could also just use `*c == '$' && c[1] != '\0'` to minimize the diff.

I'm fine either ways, but using index seemed like a better idea.

src/scrot.c
```diff
@@ -581,22 +581,24 @@ static Bool scrotXEventVisibility(Display *dpy, XEvent *ev, XPointer arg)
...
+            ch = strf[++i];
```
N-R-K, line 600: As a style thing, I'm not a fan of messing with the loop index inside the loop (mainly why I haven't merged this yet). Could use a state-machine based approach but that'll require a tad bit more code.

I'll go ahead and merge this if no one else has any objections.

guijan:
The code could just copy the '\0' from str to ret as if it were any other character, then at the bottom of the loop, break out before incrementing if it copied a '\0' like the example in https://github.com/resurrecting-open-source-projects/scrot/pull/250#issuecomment-1483060435

N-R-K:
> then at the bottom of the loop, break out before incrementing

No, because the 3rd expression in the for loop will increment `c` beyond the nul byte before the condition is assessed. Here's a small test to reproduce:

```diff
diff --git a/src/scrot.c b/src/scrot.c
index 651c5d0..e381482 100644
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -600,6 +600,7 @@ static char *imPrintf(char *str, struct tm *tm, char *filenameIM,
 
     if (strftime(strf, 4095, str, tm) == 0)
         errx(EXIT_FAILURE, "strftime returned 0");
+    strf[strlen(strf) + 1] = 'X';
 
     imlib_context_set_image(im);
     for (c = strf; *c != '\0'; c++) {
```

Now run `scrot file$` under a debugger and you'll notice that it starts reading the `X` past the string.

guijan:
I was thinking of something like this. How about it?
```c
static char *imPrintf(char *str, struct tm *tm, char *filenameIM,
    char *filenameThumb, Imlib_Image im)
{
    char *c;
    char buf[20];
    Stream ret = {0};
    long hostNameMax = 0;
    char strf[4096];
    char *tmp;
    struct stat st;

    if (strftime(strf, 4095, str, tm) == 0)
        errx(EXIT_FAILURE, "strftime returned 0");

    imlib_context_set_image(im);
    for (c = strf;; c++) {
        switch(*c) {
        case '$':
            c++;
            switch (*c) {
            case 'a':
                /* freebsd and macos don't have HOST_NAME_MAX defined.
                 * instead sysconf is recommended by freebsd.
                 * +1 for nul-terminator. */
                if (hostNameMax == 0)
                    hostNameMax = sysconf(_SC_HOST_NAME_MAX) + 1;

                streamReserve(&ret, hostNameMax);
                char *target = ret.buf + ret.off;
                gethostname(target, hostNameMax);
                ret.off += strlen(target);
                assert(ret.buf[ret.off] == '\0');
                break;
            case 'f':
                if (filenameIM)
                    streamStr(&ret, filenameIM);
                break;
            case 'm': /* t was already taken, so m as in mini */
                if (filenameThumb)
                    streamStr(&ret, filenameThumb);
                break;
            case 'n':
                if (filenameIM) {
                    tmp = strrchr(filenameIM, '/');
                    streamStr(&ret, tmp ? tmp+1 : filenameIM);
                }
                break;
            case 'w':
                snprintf(buf, sizeof(buf), "%d", imlib_image_get_width());
                streamStr(&ret, buf);
                break;
            case 'h':
                snprintf(buf, sizeof(buf), "%d", imlib_image_get_height());
                streamStr(&ret, buf);
                break;
            case 's':
                if (filenameIM) {
                    if (!stat(filenameIM, &st)) {
                        int size;

                        size = st.st_size;
                        snprintf(buf, sizeof(buf), "%d", size);
                        streamStr(&ret, buf);
                    } else
                        streamStr(&ret, "[err]");
                }
                break;
            case 'p':
                snprintf(buf, sizeof(buf), "%d",
                    imlib_image_get_width() * imlib_image_get_height());
                streamStr(&ret, buf);
                break;
            case 't':
                tmp = imlib_image_format();
                if (tmp)
                    streamStr(&ret, tmp);
                break;
            case 'W':
                if (clientWindow) {
                    if (!(tmp = scrotGetWindowName(clientWindow)))
                        break;
                    streamStr(&ret, tmp);
                    free(tmp);
                }
                break;
            default:
                goto writechar;
            }
            break;
        case '\\':
            switch (c[1]) {
            case 'n':
                c[1] = '\n';
                /* FALLTHROUGH */
            case '\\':
                c++;
                break;
            }
            /* FALLTHROUGH */
        default:
writechar:
            streamChar(&ret, *c);
            if (*c == '\0')
                return ret.buf;
        }
    }
    /* NOTREACHED. */
}
```
Haven't tested it.

src/scrot.c
```diff
@@ -590,20 +590,20 @@ static Bool scrotXEventVisibility(Display *dpy, XEvent *ev, XPointer arg)
...
-    char *c;
     char buf[20];
     Stream ret = {0};
     long hostNameMax = 0;
     char strf[4096];
     char *tmp;
     struct stat st;
 
-    if (strftime(strf, 4095, str, tm) == 0)
+    const size_t strfLen = strftime(strf, sizeof(strf), str, tm);
```
N-R-K, line 600: note: Replaced the hardcoded 4095 with `sizeof(strf)`.

Previously it was one less than sizeof(strf), but there was real reason for that. `strftime` will return 0 if the resulting string (including the nul-byte) was too big.

N-R-K:
> How about it?

Not a fan of what's going on with those fallthroughs in the `'\\'` case.

PR #256, opened 2023-04-04
fix compilation on CentOS 7

### LINKED ISSUE #225: 1.8.1 not compiling in C99 mode on CentOS 7
pkgsrc's automated testing says 1.8.1 fails to build on CentOS 7:
https://us-central.manta.mnx.io/pkgsrc/public/reports/Linux/el7/trunk/x86_64/20230315.2239//scrot-1.8.1/build.log
https://bulktracker.appspot.com/pkg/ag1zfmJ1bGt0cmFja2VyciILEgVidWlsZBiAgICD5MX3CwwLEgNwa2cYgICAvfWc1AoM

The error message is:
```text
scrot.c:346:13: error: 'for' loop initial declarations are only allowed in C99 mode
```
For whatever reason it's compiling in C89 mode. Something wrong with our configure.ac?

guijan:
I haven't actually tested it on a CentOS 7 system because I don't have a spare system to try it on and OpenBSD's virtualization daemon is very limited, the daemon doesn't work with every Linux distribution.

Should fix #225.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -10,6 +10,11 @@ AC_CONFIG_HEADERS([src/config.h])
 # Checks for programs.
 orig_CFLAGS="${CFLAGS}" # Save CFLAGS before AC_PROG_CC sets them.
 AC_PROG_CC
+# In autoconf < 2.70, AC_PROG_CC determines the compiler, and AC_PROG_CC_STDC
+# sets it to the newest C standard supported by autoconf.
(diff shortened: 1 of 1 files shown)


N-R-K (commented):
I don't have any easy way to test this either, so unless someone has anything to say, I'll merge this in a few days.

N-R-K:
Thanks.

PR #257, opened 2023-04-04
getopt_long() style cleanup

guijan:
Alphabetically sort getopt_long()'s optstring, struct option array (by the short options), and switch statement's cases.

Align the text of all the members in our struct option array's initializer.

Remove 2 useless cases. getopt_long() will never return 0 with our usage, and if it does, the behavior of the 0 case isn't what we want. Also change the '?' case to default so scrot hard errors in the unlikelihood that our assumptions are broken, and remove the duplicate (and previously useless because it never happened) default.

Remove an useless comment (/* actions */).

Sort variable declarations by order of use.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -348,57 +348,59 @@ static const char *getPathOfStdout(void)
 
 void optionsParse(int argc, char *argv[])
 {
-    static char stropts[] = "a:ofipbcd:e:hmq:s::t:uvzn:l:D:k::C:S:F:M:w:";
-
+    static char stropts[] = "a:bC:cD:d:e:F:fhik::l:M:mn:opq:S:s::t:uvw:z";
     static struct option lopts[] = {
-        /* actions */
-        { "help", no_argument, 0, 'h' },
-        { "version", no_argument, 0, 'v' },
-        { "count", no_argument, 0, 'c' },
-        { "focused", no_argument, 0, 'u' },
-        { "focussed", no_argument, 0, 'u' }, /* macquarie dictionary has both spellings */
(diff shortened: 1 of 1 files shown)


src/options.c
```diff
@@ -349,58 +349,60 @@ static const char *getPathOfStdout(void)
 
 void optionsParse(int argc, char *argv[])
 {
-    static char stropts[] = "a:ofipbcd:e:hmq:s::t:uvzn:l:D:k::C:S:F:M:w:";
-
+    int optch;
```
N-R-K, line 352: Any specific reason for bringing this above the static arrays? Don't really like mixing static and automatic variables.

src/options.c
```diff
@@ -349,58 +349,60 @@ static const char *getPathOfStdout(void)
...
+        {"script",          required_argument,  NULL,   'S'},
+        {"select",          optional_argument,  NULL,   's'},
+        {"thumb",           required_argument,  NULL,   't'},
+        {"focused",         no_argument,        NULL,   'u'},
+        /* macquarie dictionary has both spellings */
+        {"focussed",        no_argument,        NULL,   'u'},
+        {"version",         no_argument,        NULL,   'v'},
+        {"window",          required_argument,  NULL,   'w'},
+        {"silent",          no_argument,        NULL,   'z'},
+        {NULL,              0,                  NULL,    0 }
```
N-R-K, line 383: I prefer just `{0}`. But it's your pick - I'm fine either ways.

src/options.c
guijan: No, will revert.

N-R-K (approved):

PR #259, opened 2023-04-05
optionsNameThumbnail: fix sign integer warning

N-R-K:
fixes a warning in the ubuntu CI

	../../../src/options.c:542:47: warning: operand of ‘?:’ changes signedness from ‘long int’ to ‘size_t’ {aka ‘long unsigned int’} due to unsignedness of other operand [-Wsign-compare]
	  542 |     const size_t baseNameLength = extension ? extension-name : nameLength;

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -535,11 +535,11 @@ static void showVersion(void)
 
 static char *optionsNameThumbnail(const char *name)
 {
-    const size_t nameLength = strlen(name);
+    const ptrdiff_t nameLength = strlen(name);
     const char thumbSuffix[] = "-thumb";
     Stream ret = {0};
     const char *const extension = strrchr(name, '.');
-    const size_t baseNameLength = extension ? extension-name : nameLength;
+    const ptrdiff_t baseNameLength = extension ? extension-name : nameLength;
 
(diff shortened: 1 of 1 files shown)


N-R-K:
Was not enabling `-Werror` in the CI a deliberate choice or just an oversight?

guijan:
Oversight.

src/options.c
```diff
@@ -535,11 +535,11 @@ static void showVersion(void)
 
 static char *optionsNameThumbnail(const char *name)
 {
-    const size_t nameLength = strlen(name);
+    const ptrdiff_t nameLength = (ptrdiff_t)strlen(name);
```
guijan, line 538: Let the assignment do the equivalent of a cast unless that also produces a warning. It's the same issue as using the type and not the variable name in sizeof's operand.

N-R-K: > Let the assignment do the equivalent of a cast unless that also produces a warning.

Doesn't seem to produce a warning as for now at least. (and if it did, then we'd probably just have to switch to size_t (since that's that the `stream*` functions accept as param) or just disable the warning).

PR #261, opened 2023-04-05
improve timing in `scrot --delay`

guijan:
This will conflict with #257, has only been minimally tested, and there are other places in the code where it's useful but hasn't been put to use yet, so it remains a draft.

Old `nanosleep()` code (master) on OpenBSD:
openbsd-scrot-nanosleep
New `clock_nanosleep()` code (this PR), falling back to a local `clock_nanosleep()` implementation on systems lacking this POSIX function like OpenBSD and OS X, running on OpenBSD: 
openbsd-scrot-clock_nanosleep
The timing error is consistently smaller across several runs. If you subtract the user and system runtime and the 60 second sleep from the real runtime, you get the timing error. It has been reduced by an order of magnitude here.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -380,8 +380,8 @@ void optionsParse(int argc, char *argv[])
             opt.display = optarg;
             break;
         case 'd':
-            opt.delay_selection = *optarg == 'b';
-            if (opt.delay_selection)
+            opt.delaySelection = *optarg == 'b';
+            if (opt.delaySelection)
                 ++optarg;
             opt.delay = optionsParseNum(optarg, 0, INT_MAX, &errmsg);
             if (errmsg) {
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -75,6 +75,7 @@ static Imlib_Image scrotGrabFocused(void);
 static void applyFilterIfRequired(void);
 static Bool scrotXEventVisibility(Display *, XEvent *, XPointer);
 static Imlib_Image scrotGrabAutoselect(void);
+static void scrotWaitUntil(const struct timespec *);
 static Imlib_Image scrotGrabShotMulti(void);
 static Imlib_Image scrotGrabShotMonitor(void);
 static Imlib_Image scrotGrabStackWindows(void);
@@ -101,6 +102,9 @@ int main(int argc, char *argv[])
     char *haveExtension = NULL;
     struct timespec timeStamp;
     struct tm *tm;
+
+    /* Get the time ASAP to reduce the timing error in case --delay is used. */
+    clock_gettime(CONTINUOUS_CLOCK, &opt.delayStart);
 
     atexit(uninitXAndImlib);
 
@@ -285,33 +289,46 @@ static Imlib_Image scrotGrabAutoselect(void)
     return im;
 }
 
-/* similar to sleep, but deals with EINTR.
- * ideally, we'd want to sleep against an absolute time to prevent any drift
- * but clock_nanosleep doesn't seem to be widely implemented.
- */
-static void scrotSleep(int sec)
-{
-    assert(sec > 0);
-    struct timespec delay = { .tv_sec = sec };
-    while (nanosleep(&delay, &delay) < 0 && errno == EINTR);
-}
-
 void scrotDoDelay(void)
 {
-    if (opt.delay) {
-        if (opt.countdown) {
-            int i;
-
-            dprintf(STDERR_FILENO, "Taking shot in %d.. ", opt.delay);
-            scrotSleep(1);
-            for (i = opt.delay - 1; i > 0; i--) {
-                dprintf(STDERR_FILENO, "%d.. ", i);
-                scrotSleep(1);
-            }
-            dprintf(STDERR_FILENO, "0.\n");
-        } else
-            scrotSleep(opt.delay);
+    if (!opt.delay)
+        return;
+    if (opt.countdown) {
+        dprintf(STDERR_FILENO, "Taking shot in ");
+        for (int i = opt.delay; i > 0; i--) {
+            dprintf(STDERR_FILENO, "%d.. ", i);
+            opt.delayStart.tv_sec++;
+            scrotWaitUntil(&opt.delayStart);
+        }
+        dprintf(STDERR_FILENO, "0.\n");
+    } else {
+        opt.delayStart.tv_sec += opt.delay;
+        scrotWaitUntil(&opt.delayStart);
     }
+}
+
+/* scrotWaitUntil: clock_nanosleep with a simpler interface and no EINTR nagging
+ */
+static void scrotWaitUntil(const struct timespec *time)
+{
+    /* OpenBSD and OS X lack clock_nanosleep(), so we use a trivial algorithm to
+     * correct for drift and call nanosleep() everywhere.
+     */
+    struct timespec tmp;
+    do {
+        clock_gettime(CONTINUOUS_CLOCK, &tmp);
+
+        /* XXX: Use timespecsub(). OS X doesn't have that BSD macro, and libbsd
+         * doesn't support OS X save for an unmaintained fork. libobsd supports
+         * OS X but doesn't have the macro yet.
+         */
+        tmp.tv_sec = time->tv_sec - tmp.tv_sec;
+        tmp.tv_nsec = time->tv_nsec - tmp.tv_nsec;
+        if (tmp.tv_nsec < 0) {
+            tmp.tv_sec--;
+            tmp.tv_nsec += 1000000000L;
+        }
+    } while (nanosleep(&tmp, &tmp) == -1 && errno == EINTR);
 }
 
 /* Clip rectangle nicely */
--- a/src/util.h
+++ b/src/util.h
@@ -1,6 +1,6 @@
 /* util.h
 
-Copyright 2021 Guilherme Janczak <guilherme.janczak@yandex.com>
+Copyright 2021,2023 Guilherme Janczak <guilherme.janczak@yandex.com>
 Copyright 2023 NRK <nrk@disroot.org>
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
@@ -25,6 +25,19 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 */
 
 #pragma once
+
+#include <time.h>
+
+/* On Linux, CLOCK_MONOTONIC does not progress while the system is suspended,
+ * and an alternative non-standard clock which does not suffer from this problem
+ * called CLOCK_BOOTTIME is available. Scrot's CONTINUOUS_CLOCK has the exact
+ * same semantics as CLOCK_MONOTONIC, only it avoids this bug.
+ */
+#if defined(__linux__)
+    #define CONTINUOUS_CLOCK CLOCK_BOOTTIME
+#else
+    #define CONTINUOUS_CLOCK CLOCK_MONOTONIC
+#endif
 
 #define ARRAY_COUNT(X)   (sizeof(X) / sizeof(0[X]))
 #define MAX(A, B)        ((A) > (B) ? (A) : (B))
--- a/src/options.h
+++ b/src/options.h
@@ -38,6 +38,8 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 #pragma once
 
+#include <time.h>
+
 #include "scrot_selection.h"
 
 // General purpose enum
@@ -49,6 +51,7 @@ enum Direction {
 
 struct ScrotOptions {
     int delay;
+    struct timespec delayStart;
     int quality;
     enum thumb { THUMB_DISABLED, THUMB_PERCENT, THUMB_RES } thumb;
     int thumbPercent;
@@ -75,7 +78,7 @@ struct ScrotOptions {
     int autoselectW;
     SelectionMode selection;
     int monitor;
-    bool delay_selection;
+    bool delaySelection;
     bool countdown;
     bool border;
     bool silent;
--- a/src/scrot_selection.c
+++ b/src/scrot_selection.c
@@ -428,8 +428,7 @@ Imlib_Image scrotSelectionSelectMode(void)
 
     opt.selection.mode = SELECTION_MODE_CAPTURE;
 
-    /* if --delay-select is active, then do the delay before selection */
-    if (opt.delay_selection)
+    if (opt.delaySelection)
         scrotDoDelay();
 
     if (!scrotSelectionGetUserSel(&rect0))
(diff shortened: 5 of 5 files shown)


guijan:
Turns out the other `nanosleep()` calls shouldn't even be there (#274), so this PR is now ready for review now that it's rebased and tested on OpenBSD.

src/util.c
```diff
@@ -83,3 +83,43 @@ void streamStr(Stream *buf, const char *str)
...
+
+/* clock_nanosleep: clock_nanosleep for backwards systems lacking this function
+ *
+ * POSIX says that if clock is CLOCK_REALTIME and CLOCK_REALTIME is set while
+ * clock_nanosleep() waits on an absolute time, the new value should determine
+ * the moment clock_nanosleep returns. This implementation does not comply with
+ * that.
+ */
+int
+clock_nanosleep(clockid_t clock, int flags, const struct timespec *timeout,
```
N-R-K, line 98: Shadowing a standard function with an incomplete implementation seems like a bad idea. Instead I'd just implement it minimally inside a `#ifdef` in `scrotWaitUntil()`. (I'm not a fan of `#ifdef`s inside of functions but it's the lesser of two evil here).

N-R-K (changes requested):

src/util.c
N-R-K: P.S: Another alternative:

```c
#if defined(HAVE_CLOCK_NANOSLEEP)
scrotWaitUntil() { ... }
#else
scrotWaitUntil() { ... } // fallback implementation
#endif
```

guijan: That's a deliberate design choice, I think it makes it easier to understand the code. 

Anyone who knows the standard `clock_nanosleep()` can understand what `scrotWaitUntil()` is doing without having to reason about `#ifdef`. Someone who wants to go deeper into the code can read my fallback `clock_nanosleep()` implementation which is compliant save for the case where the call is `clock_nanosleep(CLOCK_REALTIME, TIMER_ABSTIME, ...)`, but if the platforms doesn't have `clock_nanosleep()`, it most likely also doesn't have an equivalent function to use (OpenBSD does have  `__thrsleep()` but it's a private API), so that's a bug we can't avoid on those platforms if we were to add a sleep on CLOCK_REALTIME to scrot anyway.

Can't it stay as is?

N-R-K: > I think it makes it easier to understand the code

It also adds unnecessary footgun into the codebase because people will need to be aware that the fallback path isn't fully compliant. No one's going to intuitively realize that we have a incomplete fallback and might end up using unimplemented feature.

N-R-K: P.S: what makes standard function "easy to understand" is that they are standard, and thus should behave predictably across systems. But the moment someone adds an incomplete fallback, that predictability is lost.

guijan: What if I make the fallback implementation call `err(1, "%s line %d: sleeping on an absolute CLOCK_REALTIME is unimplemented", __FILE__, __LINE__);` in case it's called with `clock_nanosleep(CLOCK_REALTIME, TIMER_ABSTIME, ...)`?

guijan:
Now the PR is significantly different, I've done very basic testing on OpenBSD, but it still needs more testing.

src/scrot.c
```diff
@@ -285,33 +287,55 @@ static Imlib_Image scrotGrabAutoselect(void)
...
-    assert(sec > 0);
-    struct timespec delay = { .tv_sec = sec };
-    while (nanosleep(&delay, &delay) < 0 && errno == EINTR);
+    if (!opt.delay)
+        return;
+    if (opt.countdown) {
+        int i;
+
+        dprintf(STDERR_FILENO, "Taking shot in ");
+        for (i = opt.delay; i > 0; i--) {
```
N-R-K, line 298: I'd declare `i` inside the for loop. We already do it in other places.

```suggestion
        for (int i = opt.delay; i > 0; i--) {
```

src/options.c
```diff
@@ -380,9 +382,15 @@ void optionsParse(int argc, char *argv[])
             opt.display = optarg;
             break;
         case 'd':
-            opt.delay_selection = *optarg == 'b';
-            if (opt.delay_selection)
+            if (!delayStartSet) {
+                clock_gettime(CLOCK_MONOTONIC, &opt.delayStart);
```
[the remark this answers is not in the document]
guijan: I don't like that in such a case the program would get the time and do nothing with it unless the -d flag is passed. The program's logic wouldn't follow what it's actually doing.

Although at least it's less lines of code.

src/scrot.c
```diff
@@ -285,33 +287,55 @@ static Imlib_Image scrotGrabAutoselect(void)
...
+#if defined(HAVE_CLOCK_NANOSLEEP)
+    while (clock_nanosleep(CLOCK_MONOTONIC, TIMER_ABSTIME, time, NULL)
+        == -1 && errno == EINTR);
+#else
+    /* OpenBSD and OS X lack clock_nanosleep(), so we use a trivial algorithm to
+     * correct for drift and call nanosleep().
+     */
+    struct timespec tmp;
+    do {
+        clock_gettime(CLOCK_MONOTONIC, &tmp);
```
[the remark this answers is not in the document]
guijan: Never heard of a platform missing that.

Here are some platforms we don't have CI for and which support it:
https://github.com/haiku/haiku/blob/b809279cd8dd40ef5deca391e44f4bb19eb2e54f/headers/posix/time.h#L34
https://illumos.org/man/3C/clock_gettime
https://www.ibm.com/docs/en/aix/7.2?topic=c-clock-getres-clock-gettime-clock-settime-subroutine

src/options.c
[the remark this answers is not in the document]
N-R-K: > Although at least it's less lines of code.

Not just that, I'd also be more accurate, which is one of the points of this PR.

(I don't really think that, or this PR in general, would ever make a difference in practical use-cases. But if the point of the PR is to increase accuracy, then it makes sense to do that as well).

src/scrot.c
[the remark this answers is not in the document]
N-R-K: > Here are some platforms we don't have CI for and which support it:

OK, seems fine then.

src/scrot.c
```diff
@@ -285,33 +287,55 @@ static Imlib_Image scrotGrabAutoselect(void)
...
+         */
+        tmp.tv_sec = time->tv_sec - tmp.tv_sec;
+        tmp.tv_nsec = time->tv_nsec - tmp.tv_nsec;
+        if (tmp.tv_nsec < 0) {
+            tmp.tv_sec--;
+            tmp.tv_nsec += 1000000000L;
+        }
+    } while (nanosleep(&tmp, &tmp) == -1 && errno == EINTR);
+#endif
+    errno = savedErrno;
```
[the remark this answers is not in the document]
guijan: imlib has the undocumented behavior that it could set errno, but has no way of communicating that it did.

POSIX does say `The value of errno should only be examined when it is indicated to be valid by a function's return value.` so I guess this is against the POSIX style of programming, will do another force-push to change it.

src/scrot.c
```diff
@@ -285,33 +290,53 @@ static Imlib_Image scrotGrabAutoselect(void)
...
- * ideally, we'd want to sleep against an absolute time to prevent any drift
- * but clock_nanosleep doesn't seem to be widely implemented.
- */
-static void scrotSleep(int sec)
+void scrotDoDelay(void)
 {
-    assert(sec > 0);
-    struct timespec delay = { .tv_sec = sec };
-    while (nanosleep(&delay, &delay) < 0 && errno == EINTR);
+    if (!opt.delay)
```
N-R-K, line 295: This should either be `opt.delay <= 0` or `opt.delay` should be initialized to 0 (currently it's initialized to -1, which makes this check useless):

```console
$ ./scrot -c
Taking shot in 0.
```

There was no `--delay` above, but the early return never worked and so the function is executed every time.

src/scrot.c
```diff
@@ -285,33 +290,53 @@ static Imlib_Image scrotGrabAutoselect(void)
...
-                dprintf(STDERR_FILENO, "%d.. ", i);
-                scrotSleep(1);
-            }
-            dprintf(STDERR_FILENO, "0.\n");
-        } else
-            scrotSleep(opt.delay);
-    }
+    int savedErrno = errno;
+#if defined(HAVE_CLOCK_NANOSLEEP)
+    while (clock_nanosleep(CLOCK_MONOTONIC, TIMER_ABSTIME, time, NULL)
```
N-R-K, line 317: One more thing: `MONOTONIC` doesn't count time when the system is suspended. `CLOCK_BOOTTIME` does, but that seems like it's linux specific and non-POSIX.

N-R-K (changes requested):

src/scrot.c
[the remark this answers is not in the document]
N-R-K: > That's probably why the original author of that line used `err()`.

Yeah, it was just working by accident. Imlib2's way of communicating error is via the error return (and in newer version, `imlib_get_error()`).

src/scrot.c
guijan: Interesting, CLOCK_BOOTTIME is also on OpenBSD. Will write a test program and see if it has the same caveat.

N-R-K: FreeBSD also has it: https://man.freebsd.org/cgi/man.cgi?query=clock_gettime&manpath=FreeBSD+13.2-RELEASE+and+Ports

Doesn't mention anything about what happens during system suspension though.

> Guess we'll need #ifdef __linux__ to work around that Linux bug.

If BOOTTIME is portable enough then we should just use it.

N-R-K: >If BOOTTIME is portable enough then we should just use it.

Looks like netbsd, haiku etc doesn't have it. Perhaps:

```c
#ifdef CLOCK_BOOTTIME
    #define SCROT_CLOCK CLOCK_BOOTTIME
#else
    #define SCROT_CLOCK CLOCK_MONOTONIC
#endif
```

src/scrot.c
```diff
@@ -285,33 +290,51 @@ static Imlib_Image scrotGrabAutoselect(void)
...
-            dprintf(STDERR_FILENO, "0.\n");
-        } else
-            scrotSleep(opt.delay);
-    }
+#if defined(HAVE_CLOCK_NANOSLEEP)
+    while (clock_nanosleep(CLOCK_MONOTONIC, TIMER_ABSTIME, time, NULL)
+        == -1 && errno == EINTR);
+#else
+    /* OpenBSD and OS X lack clock_nanosleep(), so we use a trivial algorithm to
+     * correct for drift and call nanosleep().
```
N-R-K, line 320:
```suggestion
     * reduce drift and call nanosleep().
```

guijan: it doesn't only reduce drift, it does correct.

src/scrot.c
guijan: I went with the standard interface everywhere except for the known broken platform.

src/scrot.c
N-R-K: Time still passes between the `clock_gettime` call and the `nanosleep` call. If unfortunate, the kernel might perform a context switch and put our process to sleep while the timespecsub is happening, which can have measurable impact (depending on the kernel config/scheduler etc).

guijan: That's true, but the same happens between `clock_nanosleep()` and anything after. 

By drift, I mean if you call `nanosleep()` in a loop, you have drift proportional to O(n) where n is the amount of times you call `nanosleep()`. I.e., if you sleep 1 second and do a task that takes 100ms in a loop, and you do that task 10 times, you will be 1 second off. Calling `clock_gettime()` and sleeping for the difference between the target time and the current time reduces that drift to O(1).

src/scrot.c
N-R-K: Are you sure that not accounting for system suspend is linux specific behavior? Because POSIX doesn't mention anything about system suspend (neither does FreeBSD manpage).

guijan: https://pubs.opengroup.org/onlinepubs/9699919799/functions/clock_gettime.html says:
```plain
If the Monotonic Clock option is supported, all implementations shall support 
a clock_id of CLOCK_MONOTONIC defined in <time.h>. This clock represents the
monotonic clock for the system. For this clock, the value returned by 
clock_gettime() represents the amount of time (in seconds and nanoseconds) 
since an unspecified point in the past (for example, system start-up time, or 
the Epoch). This point does not change after system start-up time.
```
It says amount of seconds since that point in the past, whether the machine was suspended or not does not change how much time passed. Linux [tried to fix its CLOCK_MONOTONIC](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=d6ed449afdb38f89a7b38ec50e367559e1b8f71f) once, but then [reverted the change](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=a3ed0e4393d6885b4af7ce84b437dc696490a530) because userspace applications depended on the behavior.

src/options.c
```diff
@@ -380,8 +380,9 @@ void optionsParse(int argc, char *argv[])
             opt.display = optarg;
             break;
         case 'd':
-            opt.delay_selection = *optarg == 'b';
-            if (opt.delay_selection)
+
```
N-R-K, line 383: Most other case statements don't have an immediate blank line. Was this intentional?

src/util.h
```diff
@@ -26,6 +26,19 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 #pragma once
 
+#include <time.h>
```
N-R-K, line 29: Don't think this include was needed just for the `#ifdef`, but it's fine, don't care much about it.

N-R-K (commented):
Looks mostly OK, I'll test it out sometime today.

src/util.h
guijan: Don't want the header to depend on headers it doesn't include or on the order it is included in a .c file.

N-R-K (approved):
Tested it out and it seems to work OK.

However, one weird thing I noticed while stepping through the new code under gdb was that it was taking the `nanosleep` path instead of the `clock_nanosleep` one. Which was weird since linux should support `clock_nanosleep`. Turned out, the reason was because I didn't re-run the configure scripts.

But this lead me to question how much value clock_nanosleep is providing anyways, is it worth the check and the `ifdef` ?

image

On my system at least, the answer seems to be no.

I'm fine merging this PR as-is. But if you don't have any objection to it, I'd like to just use the `nanosleep` path unconditionally and remove the `clock_nanosleep` check.

guijan:
Yeah, in an algorithmic sense, both paths are the same: they both make the error relative to the sleep constant. It just annoys me when we don't use the good features of each platform. `#ifdef` removed.

src/scrot.c
```diff
@@ -285,33 +289,46 @@ static Imlib_Image scrotGrabAutoselect(void)
...
+         * doesn't support OS X save for an unmaintained fork. libobsd supports
+         * OS X but doesn't have the macro yet.
+         */
+        tmp.tv_sec = time->tv_sec - tmp.tv_sec;
+        tmp.tv_nsec = time->tv_nsec - tmp.tv_nsec;
+        if (tmp.tv_nsec < 0) {
+            tmp.tv_sec--;
+            tmp.tv_nsec += 1000000000L;
+        }
+    } while (nanosleep(&tmp, &tmp) == -1 && errno == EINTR);
```
N-R-K, line 331: Just noticed, I don't think the 2nd `tmp` argument is doing anything. It'll just overwritten by `clock_gettime` at the top of the loop body.

PR #265, opened 2023-04-06
optionsNameThumbnail: fix edge case detecting extension

N-R-K:
if the filename ends with a dot (eg. "file.") it should not be taken as an extension.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -537,13 +537,14 @@ static char *optionsNameThumbnail(const char *name)
     const ptrdiff_t nameLength = strlen(name);
     const char thumbSuffix[] = "-thumb";
     Stream ret = {0};
-    const char *const extension = strrchr(name, '.');
-    const ptrdiff_t baseNameLength = extension ? extension-name : nameLength;
+    const char *const ext = strrchr(name, '.');
+    const bool hasExtension = ext && ext[1] != '\0';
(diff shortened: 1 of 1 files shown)


guijan:
But if you do that it will create a file called file.-thumb and -thumb becomes an extension.

N-R-K:
> But if you do that it will create a file called _file.-thumb_ and _-thumb_ becomes an extension.

What's the benefit of `file-thumb.` over `file.-thumb` ? Proper image viewers shouldn't care about extensions. Perhaps some scripts that does extension based heuristics would break, but it'd most likely break on no extension filename ("file.") as well.

I've thought about adding `thumb-` as a prefix instead of adding it before the extension, but that will break sorted listing:

```console
# current
$ ls
a.png
a-thumb.png
# other stuff..

# with "thumb-" prefix
$ ls
a.png
# other stuff...
thumb-a.png
```

guijan:
Windows Explorer displays files without the extension by default, so the user would see 2 files named _file._ with the default configuration. Although scrot on Windows is largely there to give us more confidence scrot is portable at the moment and possibly catch bugs, I've never actually used it there to see how it works.

We should treat a trailing dot as a sort of null extension when manipulating the string and as no extension when setting the file format.

This is really fixing a non-issue...

N-R-K:
Both the alternative seems whaky for one reason or another, so I'll just leave it at it's current behavior.

PR #267, opened 2023-04-07
scrotSelectionGetUserSel() cleanup

N-R-K:
See each commit message for more details.

### DIFF
--- a/src/scrot_selection.c
+++ b/src/scrot_selection.c
@@ -238,7 +238,7 @@ void scrotSelectionSetDefaultColorLine(void)
 bool scrotSelectionGetUserSel(struct SelectionRect *selectionRect)
 {
     XEvent ev;
-    int done = 0;
+    enum { WAIT, DONE, ABORT } done = WAIT;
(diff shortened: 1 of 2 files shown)


src/scrot_selection.c
```diff
@@ -331,9 +331,10 @@ bool scrotSelectionGetUserSel(struct SelectionRect *selectionRect)
                 scrotSelectionMotionDraw(rx, ry, ev.xbutton.x, ev.xbutton.y);
                 break;
             }
-            case KeyRelease:
-                /* ignore */
+            case DestroyNotify:
+                errx(EXIT_FAILURE, "received DestroyNotify event");
                 break;
+            case KeyRelease: /* ignore */
```
guijan: Should just remove this case because it'll show up in the binary, put a comment mentioning ignoring KeyRelease is intentional if you want to.

guijan:
Also, maybe we can write our own error handler to make xlib's error messages fall into the same style as others.

PR #269, opened 2023-04-09
OptionsParseNumBase() small cleanup

guijan:
The comment explaining its usage from when we only had OptionsParseNum(), and its usage was incorrect.

Also split up a nearby line that was over 80 chars.


guijan:
Sorry, I fumbled the commit message and had to force push.

N-R-K (approved):
I'd just remove the usage part entirely. People can just grep for it to see the usage (which is more likely to be up to date/correct since it's actually checked by the compiler unlike comments).

guijan:
I like that idea, done.

PR #279, opened 2023-05-02
manpage: --window option takes an argument

N-R-K:


man/scrot.txt
```diff
@@ -72,7 +72,7 @@ OPTIONS
                             500x200, 100x0, 0x480.
   -u, --focused             Use the currently focused window.
   -v, --version             Output version information and exit.
-  -w, --window              Window identifier to capture.
+  -w, --window WID          Window identifier to capture.
                             Must be a valid identifier (see xwininfo(1)).
```
N-R-K, line 76: Maybe this should be changed to:

```suggestion
                            WID must be a valid identifier (see xwininfo(1)).
```

PR #282, opened 2023-05-03
Line mode improvements

N-R-K:
### fix: loose suboption parsing for --line

currently, it unnecessarily uses strncmp and thus ends up accepting
"edgee" as "edge" etc.

### use an enum for internal line mode

no need to carry around a string internally. removes unnecessary
allocation and string comparisons.


guijan (approved):
I like removing unnecessary use of strings and the preprocessor.

PR #288, opened 2023-05-04
Remove unnecessary string duplication

N-R-K:
There are many cases in scrot where strings are unnecessarily duplicated. The goal of this patchset is to remove these.

Most of these strings also were never free-ed, so this also technically fixes some memory leaks (although in practice this isn't that big of an issue since scrot is a "one shot" program, not a long running process).

- - -

A couple other cases worth investigating:

```console
src/note.c:287:    return strndup(begin, length);
src/options.c:617:    opt.note = estrdup(optarg);
```

Also note: `opt.lineMode` is handled in #282.

### DIFF
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -676,11 +676,9 @@ static char *imPrintf(const char *str, struct tm *tm, const char *filenameIM,
                 streamChar(&ret, '$');
                 break;
             case 'W':
-                if (clientWindow) {
-                    if (!(tmp = scrotGetWindowName(clientWindow)))
-                        break;
+                if (clientWindow && (tmp = scrotGetWindowName(clientWindow))) {
                     streamStr(&ret, tmp);
-                    free(tmp);
+                    XFree(tmp);
                 }
                 break;
             default:
@@ -706,6 +704,7 @@ static char *imPrintf(const char *str, struct tm *tm, const char *filenameIM,
     return ret.buf;
 }
 
+/* return value should be freed by XFree() */
 static char *scrotGetWindowName(Window window)
 {
     assert(disp != NULL);
@@ -725,9 +724,8 @@ static char *scrotGetWindowName(Window window)
             &clsHint);
 
     if (status != 0) {
-        windowName = estrdup(clsHint.res_class);
+        windowName = clsHint.res_class;
         XFree(clsHint.res_name);
-        XFree(clsHint.res_class);
     }
     return windowName;
 }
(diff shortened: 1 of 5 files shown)


src/scrot.c
```diff
@@ -726,9 +725,8 @@ static char *scrotGetWindowName(Window window)
             &clsHint);
 
     if (status != 0) {
-        windowName = estrdup(clsHint.res_class);
+        windowName = clsHint.res_class;
```
N-R-K, line 728: NOTE: A bit weird that the function name is `GetWindowName` but it's returning the *class* instead.

N-R-K:
This PR is accumulating some unrelated commits as well. So I think I'll just mark this ready for review for now - the other two instances can be investigated and fixed in another PR.

N-R-K:
@guijan A review would be useful.

(For some reason github won't let me use the "request review" feature to request a review from you).

guijan:
Probably because I'm not a collaborator. Will have a look tomorrow.

src/scrot_selection.c
```diff
@@ -219,8 +219,7 @@ Status scrotSelectionCreateNamedColor(const char *nameColor, XColor *color)
 
 void scrotSelectionGetLineColor(XColor *color)
 {
-    scrotSelectionSetDefaultColorLine();
-
+    assert(opt.lineColor);
     const Status ret = scrotSelectionCreateNamedColor(opt.lineColor, color);
```
guijan, line 223: `scrotSelectionCreateNamedColor()` already `assert()`s that `opt.lineColor` (passed to it as an argument) is not NULL. We're devolving into a situation where the same condition is redundantly checked everywhere. 

Check for invalid values once where they might be acquired, not in callees. Also, does `XAllocNamedColor()` return error for us if we give it null pointers? If so, there would be no point to the `assert()` even for someone who likes it.

guijan:
Incredible how in master `opt.exec` gets `dup()`'d for no reason, then the `imPrintf()` call in `scrotExecapp()` `dup()`s it again, and then both leak... Scrot is one of those code bases with about as many bugs as lines of code.

This touches some of the code I'm least familiar with, namely X programming, so I can't vouch for the PR, but it looks fine.

src/scrot_selection.c
N-R-K: Mainly added it to be somewhat compatible with the older code. But yeah, I'll remove it.

And even in general, null checks as asserts are not that useful because if something dereferences a NULL pointer - it will typically crash (a loud signal that something's gone wrong).

asserts are more useful for checking invariants which wouldn't otherwise manifest itself loudly (and the program would silently march on forward - moving further away from the actual location where the bug occurred).

N-R-K:
Rebased to solve conflict. Only other change is removing the `assert` in `scrotSelectionGetLineColor`.

Issue #277, opened 2023-04-28
quality wrong at man page


Disonantemus:
man scrot.1
~~~
       -q, --quality NUM
              NUM must be between 1 and 100. For lossless output formats, a higher value represents better
              but  slower  compression. For lossy output formats, a higher value represents higher quality
              and larger file size. Default: 75.
~~~
It's inverse of that for lossless, saving a lossless format like default `png`:
~~~
> time scrot --quality 1; time scrot --quality 100
~~~
---

quality|size|speed
:-----:|---:|----:
1      |314kb  | 397ms
100    |5.94MB | 107ms

---

Also, using '--format webp' it's better than `png`to get smaller lossless screenshots, only `--quality 100` appears lossless to me (webp has lossless and lossy compression) you can test with 'cwebp' (libwebp), and ti would be better to have that quality control, maybe a '--lossless`option; with same file tested before:
`cwebp in.png -o out.webp -z 9`  

format|size
:-----:|---:
webp -z 9   |136kb
optipng -o7 |254kb

N-R-K:
Confirmed.

However, it's hard to tell how this is supposed to work because imlib2's "quality" thing is entirely undocumented (as far as I can see). So I'm not even sure what that manpage description is based on.

Looks like it got introduced at https://github.com/resurrecting-open-source-projects/scrot/commit/fc27d9f7301df1c7577be9080da7fc14e9c5836e. Before that commit it used to say: 

```
  -q, --quality NUM  Image  quality (1-100) high value means high size, low
                     compression. Default: 75. (Effect differs depending on
                     file format chosen).
```

Which was correct for PNG files, but *probably* not correct for other files.

Note to self: open an issue at imlib2 asking for documentation/clarification on how the "quality" value is supposed to work.

N-R-K:
> Also, using '--format webp' it's better than `png`to get smaller lossless screenshots, only `--quality 100` appears lossless to me (webp has lossless and lossy compression) you can test with 'cwebp' (libwebp), and ti would be better to have that quality control, maybe a `--lossless` option; with same file tested before: `cwebp in.png -o out.webp -z 9`

There's plan to add a separate --compression flag, see #290. So once that's implemented, you should be able to do something like `scrot --format webp --quality 100 --compression 9` to get highly compressed lossless webp (or jxl with new imlib2 version).

But until that's implemented, you can use the following, which will run the cwebp command on the new screenshot:

```console
[...]

PR #289, opened 2023-05-07
manpage: correct description of quality

N-R-K:
the --quality description was changed in fc27d9f but that description is not correct for lossless formats. correct the manpage description to reflect what actually happens.

Closes: https://github.com/resurrecting-open-source-projects/scrot/issues/277

### DIFF
--- a/man/scrot.txt
+++ b/man/scrot.txt
@@ -53,11 +53,12 @@ OPTIONS
   -o, --overwrite           By default scrot does not overwrite the output
                             FILE, use this option to enable it.
   -p, --pointer             Capture the mouse pointer.
-  -q, --quality NUM         NUM must be between 1 and 100. For lossless output
-                            formats, a higher value represents better but slower
-                            compression. For lossy output formats, a higher
-                            value represents higher quality and larger
-                            file size. Default: 75.
+  -q, --quality NUM         NUM must be between 1 and 100. A higher value
+                            represents better quality image at the cost of
+                            higher file size. A lower value will provide smaller
+                            file size at the cost of image quality. For lossless
+                            formats (such as PNG) only file size is affected.
+                            Default: 75.
   -s, --select[=OPTS]       Interactively select a window or rectangle with the
                             mouse, use the arrow keys to resize. See the -l and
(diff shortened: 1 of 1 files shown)


guijan (approved):

N-R-K:
Couple notes:

1. Aside from PNG, most other formats don't have much time difference between quality 1 and 100 (tested with PNG, jpg, webp and tiff) to warrant mentioning it in the manpage.
2. It might be worth it to add a separate --compression flag because quality and compression can be two separate thing on certain formats (such as jxl). (Also worth noting that if "compression" is set then the [PNG saver will ignore "quality"](https://git.enlightenment.org/old/legacy-imlib2/src/commit/c44f3a4124591e2174d2921951c48ce304693f54/src/modules/loaders/loader_png.c#L665-L685) - which makes sense since PNG is lossless).

guijan:
Yeah, using the `--quality` option to specify whether a lossless format's compression should be better and slower or worse and faster seems to be an outdated way of doing things from before we had webp and jxl. 

Perhaps we can imitate `flac` and `gzip` or let the user specify a number in the command line which describes the compression level:
This would use the slowest PNG compression level:
```sh
scrot -9 test.png
```
This would use the fastest:
```sh
scrot -0 test.png
```

Or we can imitate the `cwebp` encoder and not use the number as a command line option, but rather as an argument to an option:
```sh
scrot --compression 9 test.png
```
This has the downside that we don't seem to have a suitable letter to use as the short option. `cwebp` uses __z__, and the letter __z__ seems to be associated with compression (zstd, zip, 7zip, gzip, zlib, xz, etc) and that would be a good one, but it's taken.

Also, a jpg screenshot gets 4:2:0 chroma subsampling, that's bad for text and generally an outdated technique from back when codecs and encoders were more primitive, these days they give lower quantizers to chroma instead. I wonder if there's a way to tell imlib not to do chroma subsampling.

N-R-K:
I opened a separate issue for this just a moment ago: #290. If you can copy-paste your above comment there, for the sake of unifying the discussion, that'd be nice.

For the time being, until we have a separate compression flag, I'll merge this PR since it corrects inaccurate info.

Issue #260, opened 2023-04-05
Pick one of `int` or `bool` and stick to it


guijan:
Scrot uses a mix of the `int` and `bool` types to do boolean logic: https://github.com/resurrecting-open-source-projects/scrot/blob/c0f029e6150a743747dbdc7105d1cc4e98260b15/src/options.c#L440-L454
https://github.com/resurrecting-open-source-projects/scrot/blob/c0f029e6150a743747dbdc7105d1cc4e98260b15/src/options.c#L154-L157

Contributors should decide which one (and only one) should be used across the source code for consistency.

PR #292, opened 2023-05-14
boolify opt members

N-R-K:
use standard `bool` instead using `int` as bool.

Closes: https://github.com/resurrecting-open-source-projects/scrot/issues/260

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -368,13 +368,13 @@ void optionsParse(int argc, char *argv[])
             optionsParseAutoselect(optarg);
             break;
         case 'b':
-            opt.border = 1;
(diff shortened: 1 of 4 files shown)


guijan (approved):
The `isButtonPressed` variable remains: https://github.com/resurrecting-open-source-projects/scrot/blob/fafced405301d1815025f1b6194aeca1d25173af/src/scrot_selection.c#L234

Probably others, I don't know. Approved anyway.

N-R-K:
Good catch, added a commit fixing it. There's probably more cases hidden here and there. We can deal with them as they we find them.

PR #293, opened 2023-05-14
Add compression flag

N-R-K:


guijan:
I just noticed the man pages use "between X and Y inclusive" in some places and "between X and Y" in others. I wonder which we should use?

https://github.com/resurrecting-open-source-projects/scrot/blob/fafced405301d1815025f1b6194aeca1d25173af/man/scrot.txt#L149

https://github.com/resurrecting-open-source-projects/scrot/blob/fafced405301d1815025f1b6194aeca1d25173af/man/scrot.txt#L56

N-R-K:
That's something I've noticed too. One other option I thought about was using the mathametical notation for ranges. E.g `[0, 10]` to represent 0 to 10 inclusive.

But perhaps not everyone is familiar with such notation (though it should be very easy to figure it out just by trying `--someflag LIMIT`).

man/scrot.txt
```diff
@@ -188,6 +192,13 @@ NOTE FORMAT
     $ scrot -n "-f '/usr/share/fonts/TTF/DroidSans-Bold/40' -x 10
             -y 20 -c 255,0,0,255 -t 'Hi'"
 
+COMPRESSION QUALITY
+
+For lossless formats (e.g PNG), the quality options is ignored. For lossy
+formats where the quality and compression are tied together (e.g JPEG),
+compression will be ignored. And for image formats where quality and
+compression can be independently set (e.g WebP, JXL), both flags are respected.
```
N-R-K, line 200: NOTE: for formats that support both lossless and lossy encoding, quality of 100 typically implies lossless.

However I didn't mention it in the manpage because current imlib2 version doesn't support lossless webp (it's added in git, so will be supported next release).

N-R-K:
One more thing I realized, by setting a "compression" by default, it will break scripts where previously `--quality 0` was used as a way to get highest compression of PNG images.

Maybe we can make it so that compression is only set if user explicitly asks for it. But that makes the interaction between quality and compression more complex. And it also prevents us from setting a "good default".

guijan:
Although such scripts will no longer use the highest compression level, the program still behaves functionally the same, so I wouldn't do anything.

guijan (approved):

PR #295, opened 2023-05-18
manpage: use mathametical range notation

N-R-K:
it's shorter, and not too difficult to figure out how it works even if you're unaware of it (e.g just trying out `--flag LIMIT`).

ref: https://en.wikipedia.org/wiki/Interval_(mathematics)

### DIFF
--- a/man/scrot.txt
+++ b/man/scrot.txt
@@ -53,7 +53,7 @@ OPTIONS
   -o, --overwrite           By default scrot does not overwrite the output
                             FILE, use this option to enable it.
   -p, --pointer             Capture the mouse pointer.
-  -q, --quality NUM         NUM must be between 1 and 100. A higher value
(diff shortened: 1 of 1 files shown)


N-R-K:
> Rename the 1st commit to "manpage: use mathematical range notation". There's a typo in "mathematical".

Good catch, I'll fix it up when squashing the commits. Are the contents of the PR good or do you notice anything wrong/weird in it?

guijan:
I'd change `between [x, y]` to `in [x, y]`. A number is in a set, not between a set.

N-R-K:
> I'd change `between [x, y]` to `in [x, y]`. A number is in a set, not between a set.

I went with `within`, it should be a bit more obvious/natural to people not familiar with set-theory.

guijan (approved):

Issue #281, opened 2023-05-03
`-l opacity` does not work as documented


mimuki:
According to the man page, running a command like `scrot -s -l width=8,color="#ff79c6",opacity=255,mode=edge` should result in an opaque pink border, and should be not transparent at all.

Instead, `opacity=255` is slightly see through, but `opacity=100` *is* completely opaque.

I'm not familiar with C and might be misunderstanding this line, but it looks like this isn't a bug, just the documentation being wrong. 
https://github.com/resurrecting-open-source-projects/scrot/blob/5ef8a82612a57ebd8b70cc45eaea437cbcfed741/src/selection_edge.c#L79

N-R-K:
I can't really find any documentation on `_NET_WM_WINDOW_OPACITY` - it's not described in EWMH and seems to be something non-standard. Only thing I found was [this comment](https://github.com/tcltk/tk/blob/f7356543f35d404bef87be35b949bce5c4b168b6/unix/tkUnixWm.c#L1264) in Tk source code which suggest the atom works by mapping `[0x0, 0xFFFFFFFF]` to `[0%, 100%]` transparency.

If that's the case, then the bug I think is in the source code. Not the documentation.

N-R-K:
This will be hard for me to fix because I don't use a compositor. If you change that line to the following:

```c
	unsigned long opacity = opt.lineOpacity * (0xFFFFFFFFu / 255);
```

does that work as expected?

If my understanding of `_NET_WM_WINDOW_OPACITY` is correct, then it should work. But I'll need  help from people who can actually test it out and report back :)

daltomi:
The opacity values for the 'edge' selection mode is between 10% and 100%
At some point the function call was removed:
```
int const lineOpacity = optionsParseRequireRange(opt.lineOpacity, 10, 100);
```

N-R-K:
It was:

```c
    const int lineOpacity = optionsParseRequireRange(opt.lineOpacity,
            SELECTION_EDGE_OPACITY_MIN, SELECTION_OPACITY_MAX);
```

The `SELECTION_EDGE_OPACITY_{MIN,MAX}` constants got removed in ff88b0b0cd38137113319cb86e423cb1fe4d6b3a for some reason.

N-R-K:
> The opacity values for the 'edge' selection mode is between 10% and 100%

@daltomi What do you think should be done?

1. Change the code back to using [10, 100] for `edge` and fix the manpage.
[...]

PR #296, opened 2023-05-18
fix: incorrect line opacity mapping

N-R-K:
current code treats opacity as [0, 100] even though the manpage documents it as being [0, 255].

Fixes: https://github.com/resurrecting-open-source-projects/scrot/issues/281

### DIFF
--- a/src/selection_edge.c
+++ b/src/selection_edge.c
@@ -76,7 +76,7 @@ void selectionEdgeCreate(void)
         HeightOfScreen(scr), 0, CopyFromParent, InputOutput, CopyFromParent,
         CWOverrideRedirect | CWBackPixel, &attr);
 
-    unsigned long opacity = opt.lineOpacity * ((unsigned)-1 / 100);
+    unsigned long opacity = opt.lineOpacity * (0xFFFFFFFFu / 255);
 
     XChangeProperty(disp, pe->wndDraw,
         XInternAtom(disp, "_NET_WM_WINDOW_OPACITY", False), XA_CARDINAL, 32,


N-R-K:
It *should* fix the issue, but some actual real-world testing is still needed.

guijan:
The opacity setting doesn't seem to work in general:
```
scrot -s -l width=4,opacity=1
```
I see no difference between master and this PR, both are broken, the line isn't opaque at all, it does blend oddly with whatever is under it however.

N-R-K:
> The opacity setting doesn't seem to work in general:

Are you running under a compositor that supports `_NET_WM_WINDOW_OPACITY`?

https://github.com/resurrecting-open-source-projects/scrot/blob/6e2b5221bc25bb073552a2b0707ee99803f1a44f/man/scrot.txt#L163-L164

> since nobody complained about it not working in all this time?

It seems to be working fine (except the 100 vs 255 part) for the person that opened the bug report https://github.com/resurrecting-open-source-projects/scrot/issues/281:

> Instead, opacity=255 is slightly see through, but opacity=100 is completely opaque.

guijan:
I'm running xcompmgr, seems it doesn't. picom (a fork of xcompmgr) also doesn't it seems.

guijan:
Should probably document that it can't be any compositor, too.

N-R-K:
> picom (a fork of xcompmgr) also doesn't it seems

What WM were you running? Perhaps this flag makes it work?

https://github.com/yshui/picom/blob/3aed5599c3f73cbfa53b0249795e76ab07cf9ecd/man/picom.1.asciidoc?plain=1#L130C1-L131

guijan:
spectrwm. Still doesn't work with that flag.

PR #299, opened 2023-05-20
remove some useless assert()s

guijan:
All of the removed assert()s are checking whether pointers that are soon to be dereferenced by scrot are NULL. There are many more pointers in the source code that don't get assert()ed and any OS these days will have NULL == 0 and an unmapped 0-page that causes the program to segfault on read/write, the exception would be MMU-less systems, but we don't expect people to develop on those.

The only exception is scrotGetWindowName(), but the assert() redundantly checks if the function's argument is NULL, when its only caller already does that without assert().

### DIFF
--- a/src/note.c
+++ b/src/note.c
@@ -237,7 +237,6 @@ void scrotNoteDraw(Imlib_Image im)
     if (!im)
         return;
 
-    assert(note);
     assert(note->imFont);
 
     imlib_context_set_image(im);
@@ -256,8 +255,6 @@ void scrotNoteDraw(Imlib_Image im)
 
 static void loadFont(void)
 {
-    assert(note);
-
     note->imFont = imlib_load_font(note->font);
 
     if (!note->imFont)
@@ -266,9 +263,6 @@ static void loadFont(void)
 
 static char *parseText(char **token, char const *const end)
 {
-    assert(NULL != *token);
-    assert(NULL != end);
-
     if (**token != '\'')
         return NULL;
 
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -719,7 +719,6 @@ static char *imPrintf(const char *str, struct tm *tm, const char *filenameIM,
 static char *scrotGetWindowName(Window window)
 {
     assert(disp != NULL);
-    assert(window != None);
 
     if (window == root)
         return NULL;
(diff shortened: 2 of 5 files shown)


src/note.c
```diff
@@ -266,9 +263,6 @@ static void loadFont(void)
 
 static char *parseText(char **token, char const *const end)
 {
-    assert(NULL != *token);
-    assert(NULL != end);
```
N-R-K, line 270: `end` isn't dereferenced below.

src/scrot.c
```diff
@@ -719,7 +719,6 @@ static char *imPrintf(const char *str, struct tm *tm, const char *filenameIM,
 static char *scrotGetWindowName(Window window)
 {
     assert(disp != NULL);
-    assert(window != None);
```
N-R-K, line 722: I'd keep this.

src/selection_classic.c
```diff
@@ -31,7 +31,6 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
     Part of the code comes from the scrot.c file and maintains its authorship.
 */
 
-#include <assert.h>
```
N-R-K, line 34: Leaving a standard header included doesn't really hurt. `assert.h` isn't even that big of header file, compared to `stdlib` or `unistd` - which might measurably slow compile time down.

N-R-K (changes requested):

src/note.c
guijan: `**token` is, and then there's a line with `*token != end` below that will be an infinite loop if `end` is `NULL`

src/selection_classic.c
guijan: It's unused. We're having bugs in code without which scrot would still behave the same.

src/scrot.c
guijan: It's a NULL check:
https://gitlab.freedesktop.org/xorg/proto/xorgproto/-/blob/eb28f0378fadd0d143aad7ec16f7b91814faae9a/include/X11/X.h#L115
https://github.com/resurrecting-open-source-projects/scrot/blob/6e2b5221bc25bb073552a2b0707ee99803f1a44f/src/scrot.c#L690

N-R-K: `clientWindow` is not a pointer, window IDs are integers.

N-R-K:
Merged manually in 07d0ccf6d7326394200b5109c62665f569d27e23. Only change is I added back the `window` assert. Wanted to merge this since I was going to replace the asserts with a custom variant and wanted to avoid conflicts.

PR #301, opened 2023-05-20
deprecate --display

guijan:
We don't need to duplicate the work XOpenDisplay() already does for us. Mark --display for removal at a later date.

2023-05-20-192641_678x87_scrot

### DIFF
--- a/man/scrot.txt
+++ b/man/scrot.txt
@@ -2,8 +2,8 @@ NAME
   scrot - command line screen capture utility
 
 SYNOPSIS
-  scrot [-bcfhimopuvz] [-a X,Y,W,H] [-C NAME] [-D DISPLAY] [-d SEC] [-e CMD]
-        [-k OPT] [-l STYLE] [-M NUM] [-n OPTS] [-q NUM] [-s OPTS] [-t % | WxH]
+  scrot [-bcfhimopuvz] [-a X,Y,W,H] [-C NAME] [-d SEC] [-e CMD] [-k OPT]
+        [-l STYLE] [-M NUM] [-n OPTS] [-q NUM] [-s OPTS] [-t % | WxH]
         [-w NUM] [[-F] FILE]
 
 DESCRIPTION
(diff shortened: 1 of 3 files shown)


N-R-K:
The multimonitor code seems to use the explicit display parameter - so removing `-D` probably won't buy us much LoC.

https://github.com/resurrecting-open-source-projects/scrot/blob/6e2b5221bc25bb073552a2b0707ee99803f1a44f/src/scrot.c#L891

I don't know whether this is OK or hacky as I don't use multi-monitor setup.

Also setting an environment variable `DISPLAY` is not the same as `-D` - the exec-ed child will inherit the new `$DISPLAY` but not the one set via `-D` (not sure whether this matters in practice or not - but I wanted to make sure we're aware of the subtle changes that this introduces).

guijan:
I think the difference between `-D` and the `DISPLAY` environment variable is an unintended consequence more than a feature. Are people really going to run scrot in one X server, and then exec a command in another?
Also, because `system()` invokes a shell, the user can make the exec string a script that first calls `unset DISPLAY` for this edge case.

N-R-K:
> I think the difference between `-D` and the `DISPLAY` environment variable is an unintended consequence more than a feature. Are people really going to run scrot in one X server, and then exec a command in another?

Same. But as I said, I'm just trying to make sure we fully understand the effect of the change.

- - -

Also I looked a bit more into `scrotGrabShotMulti`'s usage of `initXAndImlib` and I don't think we'll gain a whole lot by removing `-D`. Unless you want to fix/cleanup `scrotGrabShotMulti` (but keep in mind I have no way of testing such changes).

guijan:
It'll still remove and simplify code if we keep the lines that get the display name from the environment for `scrotGrabShotMulti()`. Shouldn't be too hard. We don't have to fix that and deprecate `-D` at the same time either.

N-R-K:
> It'll still remove and simplify code if we keep the lines that get the display name from the environment for `scrotGrabShotMulti()`.

I don't think I follow, what would be removed and simplified exactly? Can you show a patch or something.

N-R-K:
Looking at it again, I don't see how removing `--display` buys us anything. And it's a straightforward flag to keep supporting, unlike `--note` for example.

Issue #272, opened 2023-04-09
undocumented behavior in scrot


guijan:
scrot has some undocumented behavior and sometimes deliberately continues with some kludge when it fails to comply with the user's request.

We can see here that if the user requests that the cursor be included in the screenshot and scrot fails to capture the cursor, it will go ahead anyway without the the cursor:
https://github.com/resurrecting-open-source-projects/scrot/blob/75664efd19d02c0d70006e902364a877bec902b9/src/scrot.c#L432-L435

Here, scrot can't make sense of the user's wishes, but marches on anyway: 
https://github.com/resurrecting-open-source-projects/scrot/blob/75664efd19d02c0d70006e902364a877bec902b9/src/options.c#L503-L504

In `imPrintf()`, if the user asks us to include the thumbnail's filename in the command to execute but there is no thumbnail, we just ignore it:
https://github.com/resurrecting-open-source-projects/scrot/blob/75664efd19d02c0d70006e902364a877bec902b9/src/scrot.c#L624-L627

We used to have an entirely undocumented feature: #216.

There are more instances in the code. We should start hard erroring if we fail to do what we were told to do by the user, and we should get rid of undocumented behaviors or adopt and document them if they're sensible like #216.

PR #302, opened 2023-05-20
remove undocumented --focussed spelling of --focused

guijan:
Fixes a small part of #272.
Wikitionary says "focused" is the common spelling in the Commonwealth and the US: https://en.wiktionary.org/wiki/focussed although "focussed" is recognized as valid.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -350,8 +350,6 @@ void optionsParse(int argc, char *argv[])
         {"select",          optional_argument,  NULL,   's'},
         {"thumb",           required_argument,  NULL,   't'},
         {"focused",         no_argument,        NULL,   'u'},
-        /* macquarie dictionary has both spellings */
-        {"focussed",        no_argument,        NULL,   'u'},
         {"version",         no_argument,        NULL,   'v'},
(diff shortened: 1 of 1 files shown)


N-R-K:
This has been in scrot since 2009, 1e90918ca2191a8322ac80ba04b2bd7e0e3c87dc (the commit date doesn't match up, I presume because it was manually imported to git from some other source control system).

I think I'd just leave it be and document it instead.

N-R-K:
> I wonder if we can steal any bug fixes from that

Probably worth investigating. Also if that fork is actively maintained, we should maybe try to unify efforts instead (unless that fork has philosophy/goal incompatible with ours).

> if we can fix our history

This will end up changing the commit hashes of pretty much the entire repository. Which will break some (a lot?) of automated setups (e.g package managers).

And even in general, a repo changing it's entire commit history overnight is incredibly suspicious from an outside perspective (especially so for package maintainers).

N-R-K:
Seems like that person wasn't planning to maintain/develop that fork: https://github.com/dreamer/scrot/issues/2#issuecomment-537611069

Issue #284, opened 2023-05-04
`scrot --select` borders caught up in the screenshot


guijan:
Someone made a Youtube video about scrot and described a bug in it:
https://yewtu.be/watch?v=aD17Rd2D27c&t=250
The gist of it is that `scrot --select` (which implicitly means `scrot --select --line mode=classic`) seems to have a bug when selecting a video that is currently being played, the selection edges lag behind. Then, when you finally take the screenshot, those edges can be caught by the screenshot.

We could simply make `--line mode=edge` the default and deprecate `--line mode=classic` to fix this. However, it would also be wise to figure out why this bug is there. It would also be wise to figure out which option is "better" and deprecate the other, maybe edge is not better. Also, the man page says edge and classic support different but overlapping customization options, so the missing functionality should be added to one or the other.

N-R-K:
According to the FAQ the edge mode doesn't work properly when a compositor is running.

https://github.com/resurrecting-open-source-projects/scrot/blob/80660e7b14c8882da92a138df0db38ee616c5298/FAQ#L30-L32

N-R-K:
I was trying to look at what slop does in it's pure X implementation and found [this comment](https://github.com/naelstrof/slop/blob/499db815ee968b21454d4aae69458549c50315bf/CMakeLists.txt#L15-L18)

```
# This should really never be disabled. The pure-X mode of slop is very expensive and buggy.
# It also doesn't work on Wayland. Though if a system is never running a compositor, or
# doesn't have OpenGL, this could remove some linking dependencies I suppose.
```

So it seems like even slop (running in pure X mode) has issues when there's a compositor running.

If I get around setting up a compositor to test things out then I can probably try and see what's going on. Whatever it is, I suspect it should be *possible* to fix (since it apparently works fine with opengl), but I don't suspect it to be easy at all.

N-R-K:
Closed this since the default is now edge when no compositor is running. The issues with compositor is tracked in #76.

PR #304, opened 2023-05-21
selectionEdgeDestroy: misc cleanups

N-R-K:
* xeventUnmap() did not check the union type before reading from it.
* unmap event should always come, no need to sleep/retry 30 times.
* remove xeventUnmap() and waitUnmapWindowNotify() entirely. the main logic is short enough to inline directly at the call site.
* rather than unmapping the window, and then destroying it, destroy it directly and wait for the DestroyNotify event. (this may help with: https://github.com/resurrecting-open-source-projects/scrot/issues/284)
* compare window ID against `None` rather than `0`. no functional difference, but using `None` is more idiomatic for X11 things.

### DIFF
--- a/src/selection_edge.c
+++ b/src/selection_edge.c
@@ -50,9 +50,6 @@ struct SelectionEdge {
     XClassHint *classHint;
 };
 
-static void waitUnmapWindowNotify(void);
-static Bool xeventUnmap(Display *, XEvent *, XPointer);
-
 void selectionEdgeCreate(void)
 {
     struct Selection *const sel = *selectionGet();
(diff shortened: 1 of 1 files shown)


N-R-K:
> (this may help with: https://github.com/resurrecting-open-source-projects/scrot/issues/284)

Just noticed that that bug happens on classic selection mode, not edge. I'll remove that part of the commit message before merging.

daltomi:
This PR breaks what you want to avoid. That the selection line does not appear in the capture.
```bash
scrot --select=hide --line mode=edge,width=4,color=red
```
2023-05-25-110926_411x290_scrot

N-R-K:
> This PR breaks what you want to avoid. That the selection line does not appear in the capture.

Weird, it's working fine on my end. The selection line doesn't appear on the shot:

scrot

N-R-K:
@daltomi A couple questions

1. Are you running a compositor?
2. Does the follow patch make any difference for you?

```diff
diff --git a/src/selection_edge.c b/src/selection_edge.c
index f0cc3c7..94ac4d1 100644
--- a/src/selection_edge.c
+++ b/src/selection_edge.c
@@ -135,6 +135,7 @@ void selectionEdgeDestroy(void)
             if (ev.type == DestroyNotify && ev.xdestroywindow.window == pe->wndDraw)
                 break;
         }
+        nanosleep(&(struct timespect){ .tv_nsec = 120 * 1000L * 1000L }, NULL);
 
         XFree(pe->classHint);
     }
```

daltomi:
Yes with composer, but it also fails without composer: if you select on a window that plays video, the drag is shown:

2023-05-25-190928_700x494_scrot

The `master` branch works fine. (I originally implemented the delay to avoid these issues, answering your second question)

N-R-K:
> I originally implemented the delay to avoid these issues, answering your second question

I see, I think I understand why this is happening a bit better now. Although we received a `DestroyNotify` event, the frame still might not have been updated. Or a compositor might be buffering frames.

I pushed a new commit adding the delay back. I think it should fix the issue. (I've also added a comment, because it's not obvious why there's a sleep there).

Issue #274, opened 2023-04-10
Raise the window more reliably


guijan:
Instead of polling for the event (with drift from calling `nanosleep()` in a loop...), this code:
https://github.com/resurrecting-open-source-projects/scrot/blob/dc65e9697c0a272029d694efddb1c7cce9e66cf6/src/scrot.c#L373-L379 
Should use `select()` or `poll()` to wait for the fd used to communicate with the X server to have some data, then use `XCheckIfEvent()` to check if the data is what we're watching for.

Edit: Is 300ms enough of a delay? 10ms is often the precision OSes offer to the userland when it comes to timers, there's probably significant drift in polling every 10ms, fixing this might expose an issue with the delay being too short, but I find it unlikely for local X servers. 
Edit 2: fixed incorrect statement about `poll()`

N-R-K:
> Edit: `poll()` doesn't have subsecond precision.

`poll`'s timeout is in miliseconds.

And there are much bigger issues with that code. I'm at the end of my awake hours, so I'll take a closer look tomorrow with fresh eyes before elaborating.

guijan:
Other places with the same pattern:
https://github.com/resurrecting-open-source-projects/scrot/blob/3b281dbf20fb4e4b8bd74c177163c441561db024/src/scrot_selection.c#L248-L256
https://github.com/resurrecting-open-source-projects/scrot/blob/3b281dbf20fb4e4b8bd74c177163c441561db024/src/selection_edge.c#L152-L158

N-R-K:
> And there are much bigger issues with that code.

1. `scrotXEventVisibility` tries to check if the window becomes visible. However, that event will *never* come. Because we never subscribed to visibility events.

```c
            XSelectInput(disp, target, FocusChangeMask);
```

This subscribes us to focus-change events, not visibility events. For visibility events, `VisibilityChangeMask` is needed. The `window` member on both visibility and focus-change event happens to line up, which is why that code happened to "work" (for some definition of work).

2. Just because we received a visibility event doesn't mean the window is visible. Visibility event can be generated due to a window becoming obscured or partially obscured as well.

3. XCheckIfEvent will only pull out the event for which `scrotXEventVisibility` returned true and leave the prior events in the queue. I don't really see why this would be needed or desired.

4. There might be a race since we call `XSelectInput` _after_ raising the window. So we might end up missing the event even though it was generated. (This one I'm not 100% sure about since xlib buffers the output).

[...]

PR #305, opened 2023-05-21
scrotGetGeometry: sleep explicitly

N-R-K:
this patch simply removes the fluff and indirection and shows what's actually going on already. since we never subscribed to visibility event, we're never going to receive it anyways and will end up sleeping the entire duration.

also reduce the sleep from ~300ms to ~160ms, which should be enough for local X server. (note: it's fine to call nanosleep in loop, it's just a fuzzy sleep, doesn't have to be exact).

ref: https://github.com/resurrecting-open-source-projects/scrot/issues/274

### DIFF
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -73,7 +73,6 @@ static void uninitXAndImlib(void);
 static size_t scrotHaveFileExtension(const char *, char **);
 static Imlib_Image scrotGrabFocused(void);
 static void applyFilterIfRequired(void);
-static Bool scrotXEventVisibility(Display *, XEvent *, XPointer);
 static Imlib_Image scrotGrabAutoselect(void);
 static void scrotWaitUntil(const struct timespec *);
 static Imlib_Image scrotGrabShotMulti(void);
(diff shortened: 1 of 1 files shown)


guijan:
The problem you will have with "fuzzy sleep" is that your average kernel will only return control to the program on a scheduler tick. When I ran Gentoo, I could pick a few steps between 100 and 1000hz scheduler ticks, but I've seen OpenWRT routers do 62hz. OpenBSD defaults to 100hz and does not recommend recompiling the kernel with a different setting. Tickless kernels can return anytime, FreeBSD and Linux support this, I don't know if FreeBSD or any Linux distros default to that. 

With 10ms ticks (100hz), if you call `nanosleep()` 5ms after your process is scheduled, you will sleep for 15ms if you get scheduled on the first after your sleep is over. This is a huge deviation, and unavoidable for a single call, but really bad in a loop. And it's going to vary a lot between platforms and hardware. It could be that it's too tight or too loose in your machine and the opposite on another. We don't actually know for how long that code is sleeping, better to just use `scrotWaitUntil()` so we can be more confident on how long we're actually sleeping for and so the timing is more consistent everywhere.

N-R-K:
There's a number of reasons I didn't go with `sleepUntil` (at least for now). As you might have noticed, I used `ref:` in the commit message to reference the issue, not `Closes:` or `Fixes:` because I don't consider the issue fixed. This is just removing the elaborate sleep with an explicit one so (future) contributors can easily see what's *actually happening*.

In other words, this is just a "cleanup" patch, not a bugfix/improvement.

(Although I just realized that this patch takes 30 nanosleep calls in a loop to a single nanosleep call in the common case - since non-fatal signals arriving during sleep is not common. That's an accidental improvement.)

The only real thing I wanted to actually change with this PR is the bringing the intended sleep duration down from 300ms (which is very noticeable) to 160ms (still noticeable, but bearable). I didn't want to go *too* low (even though on my system, the WM responds within a couple milliseconds most of the time).

- - -

But more generally, this entire thing is unsound, and we don't know *for sure* if sleeping is really the best solution or not. I see 3 possible outcomes:

1. Ideal case: we'd find out a proper way to do this that I glanced over (perhaps a bit optimistic, but I'm hoping ends up being the case).
2. Turns out sleeping *really* is the only solution, in which case I wouldn't mind switching over to `sleepUntil`
3. We drop this feature entirely.

3 might seem a bit sudden - but just today I realized that I don't know any other screenshot utility that has this behavior. `maim` (+ `slop`) definitely doesn't have this behavior, and I used to use `flameshot` ~2 years ago and don't remember it having this behavior either.

Is it even useful? The user could've just raised the window himself before screenshotting. And also there's the fact that this isn't something that even works reliably (e.g nothing happens on `dwm`).

So I feel like just dropping this feature might be a viable path as well.

guijan:
Yeah, a weird feature with bad code that doesn't even work everywhere for reasons that are outside this program. I'd remove it too.

I'd wager that the average non-programmer doesn't even know that the obscured parts of windows continue to exist and even get updated. And people probably want to know what they're screenshotting, who's going to get a window's identifier so they can pass it to scrot, then blindly screenshot it?

daltomi:
> Is it even useful? The user could've just raised the window himself before screenshotting.

Of course it is useful. That's what the `--select` + `--border` complementarity  is for: being able to click on a partially hidden window and bring it to the top of the stack.
Look in #67, when clicking on the file explorer window, it only captures its area but not its content when staying one level down in the stack.
**Do not remove this option, please.**

> @guijan Yeah, a weird feature with bad code that doesn't even work everywhere

Everything you read that you didn't do it yourself is "bad code", it's annoying.

N-R-K:
> Do not remove this option, please.

If it's useful then I don't have any plans to remove it. 👍

But in that case, we probably would like to improve the feature to work more reliably. Sending a `_NET_ACTIVE_WINDOW` request along with `XRaiseWindow` might be worthwhile since some WMs don't respond to the `XRaiseWindow`.

PR #311, opened 2023-05-23
README: pkg-config is abandoned, recommend pkgconf

guijan:
The last release of pkg-config was in 2017-03-20:
https://pkgconfig.freedesktop.org/releases/

pkgconf has silently replaced pkg-config in the wild. For instance, in Debian: https://packages.debian.org/sid/pkg-config

Let's recommend the maintained implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,10 +43,10 @@ scrot requires a few projects and libraries:
 
 - [autoconf](https://www.gnu.org/software/autoconf/autoconf.html) (build time only)
 - [autoconf-archive](https://www.gnu.org/software/autoconf-archive/) (build time only)
-- [pkg-config](https://www.freedesktop.org/wiki/Software/pkg-config/) (build time only)
+- A pkg-config implementation [(e.g. pkgconf)](https://github.com/pkgconf/pkgconf) (build time only)
 - [imlib2](https://sourceforge.net/projects/enlightenment/files/imlib2-src/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -43,7 +43,7 @@ scrot requires a few projects and libraries:
 
 - [autoconf](https://www.gnu.org/software/autoconf/autoconf.html) (build time only)
 - [autoconf-archive](https://www.gnu.org/software/autoconf-archive/) (build time only)
-- [pkg-config](https://www.freedesktop.org/wiki/Software/pkg-config/) (build time only)
+- [pkgconf](https://github.com/pkgconf/pkgconf) (build time only)
```
N-R-K, line 46: pkgconf is the most adopted pkg-config implementation since the original pkg-config died. But it's not the only option, there's [u-config](https://github.com/skeeto/u-config) and there also seems to be a [ruby implementation](https://github.com/ruby-gnome/pkg-config) (which is available on the gentoo repo). Looks like there's also a [perl version](https://github.com/PerlPkgConfig/perl-PkgConfig) (though this one isn't available on gentoo repos).

So just ask for _any_ `pkg-config` implementation, and put `pkgconf` as the example since it's the most adopted one.

```suggestion
- A `pkg-config` implementation (e.g [pkgconf](https://github.com/pkgconf/pkgconf)) (build time only)
```

N-R-K (changes requested):

PR #317, opened 2023-05-25
scrotGrabMousePointer: error out on failure

N-R-K:
This turns the previous warnings into errors instead - since we couldn't accomplish what the user asked us to.

And while working on this, I also realized that the in-place mutation wouldn't violate strict aliasing and is legal. So remove the failure point created by `malloc` entirely.

ref: https://github.com/resurrecting-open-source-projects/scrot/issues/272

### DIFF
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -448,12 +448,9 @@ Window scrotGetWindow(Display *display, Window window, int x, int y)
 void scrotGrabMousePointer(const Imlib_Image image, const int xOffset,
     const int yOffset)
 {
-    XFixesCursorImage *xcim = NULL;
-    /* Get the X cursor and the information we want. */
-    if ((xcim = XFixesGetCursorImage(disp)) == NULL) {
-        warnx("Can't get the cursor from X");
-        goto end;
-    }
+    XFixesCursorImage *xcim = XFixesGetCursorImage(disp);
+    if (!xcim)
+        errx(EXIT_FAILURE, "Can't get the cursor from X");
     const unsigned short width = xcim->width;
     const unsigned short height = xcim->height;
     const size_t pixcnt = (size_t)width*height;
@@ -467,10 +464,8 @@ void scrotGrabMousePointer(const Imlib_Image image, const int xOffset,
     }
 
     Imlib_Image imcursor = imlib_create_image_using_data(width, height, pixels);
-    if (!imcursor) {
-        warnx("Can't create cursor image");
-        goto end;
-    }
+    if (!imcursor)
+        errx(EXIT_FAILURE, "Can't create cursor image");
 
     /* Overlay the cursor into `image`. */
     const int x = (xcim->x - xcim->xhot) - xOffset;
@@ -482,8 +477,6 @@ void scrotGrabMousePointer(const Imlib_Image image, const int xOffset,
         height);
     imlib_context_set_image(imcursor);
     imlib_free_image();
-
-end:
     XFree(xcim);
 }
 
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -449,7 +449,6 @@ void scrotGrabMousePointer(const Imlib_Image image, const int xOffset,
     const int yOffset)
 {
     XFixesCursorImage *xcim = NULL;
-    uint32_t *pixels = NULL;
     /* Get the X cursor and the information we want. */
     if ((xcim = XFixesGetCursorImage(disp)) == NULL) {
         warnx("Can't get the cursor from X");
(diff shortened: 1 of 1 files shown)


daltomi:
Let's try not to place commits that do not go with what is indicated by the PR (I understand, I also did it sometimes)

src/scrot.c
```diff
@@ -448,33 +448,24 @@ Window scrotGetWindow(Display *display, Window window, int x, int y)
 void scrotGrabMousePointer(const Imlib_Image image, const int xOffset,
     const int yOffset)
 {
-    XFixesCursorImage *xcim = NULL;
-    uint32_t *pixels = NULL;
-    /* Get the X cursor and the information we want. */
-    if ((xcim = XFixesGetCursorImage(disp)) == NULL) {
```
guijan, line 454: Also worthy of note is that `xcim` gets initialized with a value that is never used, that seems to be everywhere in scrot. Searching for `= NULL` in scrot.c reveals many instances.

src/scrot.c
```diff
@@ -448,33 +448,24 @@ Window scrotGetWindow(Display *display, Window window, int x, int y)
...
-    }
+    uint32_t *pixels = (uint32_t *)xcim->pixels;
     /* XFixesCursorImage returns pixels as `unsigned long`, which is typically
      * 64bits, but imlib2 expects 32bit packed integers. */
-    for (size_t i = 0; i < pixcnt; i++)
-        pixels[i] = xcim->pixels[i];
+    if (sizeof *xcim->pixels > sizeof *pixels) {
+        for (size_t i = 0; i < pixcnt; ++i)
+            pixels[i] = xcim->pixels[i];
+    }
```
guijan, line 464: This does violate C's aliasing rules: https://www.iso-9899.info/n1570.html#6.5p7

N-R-K: Yes, I've read that and explained in the commit why there's no violation. If you think otherwise then feel free to explain in more detail.

N-R-K: To elaborate, each element of an array is an object by itself ([C99 § 3.14](https://port70.net/~nsz/c/c99/n1256.html#3.14p1)). The write into `u32[0]` only changes the effective type of that region, not the entire array, the rest of the elements still retain their old effective type because they haven't been written to, [C99 § 6.5](https://port70.net/~nsz/c/c99/n1256.html#6.5p6):

> If a value is stored into an object having no declared type through an lvalue having a type that is not a character type, then the type of the lvalue becomes the effective type of the object for that access and for subsequent accesses that do not modify the stored value.

guijan: It's perfectly valid as far as C goes that `xcim->pixels` is aligned to `unsigned long`, but not `uint32_t`. There is no requirement that `uint32_t`'s alignment requirements have to be <= `unsigned long`'s. 

Perhaps we can let it slip, [Eric S. Raymond's struct packing guide](http://www.catb.org/esr/structure-packing/#_alignment_requirements) says:
>I’ve learned something rather reassuring from working with the source code for the reference implementation of NTP. It does packet analysis by reading packets off the wire directly into memory that the rest of the code sees as a struct, relying on the assumption of minimal self-aligned padding - or zero padding in odd cases like 690x0.
>
>The interesting news is that NTP has apparently being getting away with this for decades across a very wide span of hardware, operating systems, and compilers, including not just Unixes but under Windows variants as well. This suggests that platforms with padding rules other than self-alignment are either nonexistent or confined to such specialized niches that they’re never either NTP servers or clients.

i.e. in all modern machines `_Alignof(type) == sizeof(type)`, although I know that `long double` can be an exception on i386 with its 80-bit FPU: it has `sizeof(long double) == 10` and `_Alignof(long double) == 16` in the SysV ABI.

guijan: Also, no unnecessary casts as in line 458 please.

N-R-K: > Also, no unnecessary casts as in line 458 please.

The cast wasn't there on my first iteration, but it caused a compiler warning. So I added it.

> There is no requirement that `uint32_t`'s alignment requirements have to be <= `unsigned long`'s, or that `unsigned long`'s alignment requirements have to be an integer multiple of `uint32_t`'s.

Okay, that's fair. I was only thinking of aliasing and didn't consider alignment. But if such a machine exists, I kinda doubt Imlib2 will work there, because IIRC I saw some of the Imlib2 loaders treat raw memory as struct as well - which would most likely be broken under such machines.

guijan: >it caused a compiler warning
Ugh, compilers forcing us to write C++-like C.

PR #318, opened 2023-05-25
merge imlib2.c into scrot.c

N-R-K:
the file is badly named, it says "imlib.c" but half of it is X11 related. also fix a comment "Imlib stuff" => "X11 stuff".

and since there's only a single function, just move it to scrot.c and make it `static`.


guijan:
I have only given it a cursory look, but it looks good.

Issue #320, opened 2023-05-26
Scrot Capture when VLC (fullscreen) is playing video


desmphil:
Hi,

Im using scrot over a small flask web site to monitor what is (playing) on the hdmi display.
This works fine with Chromium, or other any application running.

As soon as I start a VLC video fullscreen in display=0, then the scrot capture only return black images.
If i start VLC without the --fullscreen option, the capture works.
Any idea?

scrot command line
sudo -u pi DISPLAY=:0 scrot --overwrite -t 5 -z

vlc command line
vlc --fullscreen --no-qt-privacy-ask --no-osd --loop --random --no-audio /home/pi/Videos/video1.mp4

N-R-K:
Seems to work okay with `mpv`'s fullscreen, I don't have vlc installed. Please provide some more information about your setup as well:

1. Which scrot version?
2. Which WM are you using?
3. Is a compositor involved or not?

Also try to see if you can reproduce it with `mpv` or if it's just `vlc` that's causing issue.

DogeTheBeast:
Is this bug still reproducible ? I am unable to reproduce it on Arch linux with X11

Issue #82, opened 2021-04-09
Hide a window/windows from the screenshot


beanrocks:
I know it is possible to take screenshots of all of the windows using window compositor even though they are minimized or partially hidden, using the `-k`, `--stack` option.

Now is it possible to add a feature that does that opposite? It would take a normal screenshot (not joined) but hide a window/windows from the screenshot.

Thanks!

PR #323, opened 2023-05-26
Revise scrotWaitUntil interface

N-R-K:
The `timespec` argument for nanosleep is awkward because it uses `ns` when we want to sleep for some `ms` instead. And `scrotWaitUntil` is awkward because it uses absolute time, instead of relative time.

Add `scrotWaitFor` which uses a relative time in `ms` and uses `scrotWaitUntil` internally.

CC: @guijan

### DIFF
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -74,6 +74,7 @@ static size_t scrotHaveFileExtension(const char *, char **);
 static Imlib_Image scrotGrabFocused(void);
 static void applyFilterIfRequired(void);
 static Imlib_Image scrotGrabAutoselect(void);
+static long miliToNanoSec(int);
 static void scrotWaitUntil(const struct timespec *);
 static Imlib_Image scrotGrabShotMulti(void);
 static Imlib_Image scrotGrabShotMonitor(void);
@@ -145,7 +146,7 @@ int main(int argc, char *argv[])
      * screenshot-taking above or it will skew the timing.
      */
     clock_gettime(CLOCK_REALTIME, &timeStamp);
-    if (timeStamp.tv_nsec >= 500*1000*1000) {
+    if (timeStamp.tv_nsec >= miliToNanoSec(500)) {
         /* Round the timestamp to the nearest second. */
         timeStamp.tv_sec++;
     }
@@ -345,6 +346,11 @@ void scrotDoDelay(void)
     }
 }
 
+static long miliToNanoSec(int ms)
+{
+    return ms * 1000L * 1000L;
+}
+
 /* scrotWaitUntil: clock_nanosleep with a simpler interface and no EINTR nagging
  */
 static void scrotWaitUntil(const struct timespec *time)
@@ -364,7 +370,7 @@ static void scrotWaitUntil(const struct timespec *time)
         tmp.tv_nsec = time->tv_nsec - tmp.tv_nsec;
         if (tmp.tv_nsec < 0) {
             tmp.tv_sec--;
-            tmp.tv_nsec += 1000000000L;
+            tmp.tv_nsec += miliToNanoSec(1000);
         }
     } while (nanosleep(&tmp, NULL) == -1 && errno == EINTR);
 }
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -109,7 +109,7 @@ int main(int argc, char *argv[])
     struct tm *tm;
 
     /* Get the time ASAP to reduce the timing error in case --delay is used. */
-    clock_gettime(CONTINUOUS_CLOCK, &opt.delayStart);
+    opt.delayStart = clockNow();
 
     atexit(uninitXAndImlib);
 
@@ -351,6 +351,21 @@ static long miliToNanoSec(int ms)
     return ms * 1000L * 1000L;
 }
 
+/* clockNow() has the exact same semantics as CLOCK_MONOTONIC. Except that on
+ * Linux, CLOCK_MONOTONIC does not progress while the system is suspended, so
+ * the non-standard CLOCK_BOOTTIME is used instead to avoid this bug.
+ */
+struct timespec clockNow(void)
+{
+    struct timespec ret;
+#if defined(__linux__)
+    clock_gettime(CLOCK_BOOTTIME, &ret);
+#else
+    clock_gettime(CLOCK_MONOTONIC, &ret);
+#endif
+    return ret;
+}
+
 /* scrotWaitUntil: clock_nanosleep with a simpler interface and no EINTR nagging
  */
 static void scrotWaitUntil(const struct timespec *time)
@@ -360,7 +375,7 @@ static void scrotWaitUntil(const struct timespec *time)
      */
     struct timespec tmp;
     do {
-        clock_gettime(CONTINUOUS_CLOCK, &tmp);
+        tmp = clockNow();
 
         /* XXX: Use timespecsub(). OS X doesn't have that BSD macro, and libbsd
          * doesn't support OS X save for an unmaintained fork. libobsd supports
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -75,7 +75,6 @@ static Imlib_Image scrotGrabFocused(void);
 static void applyFilterIfRequired(void);
 static Imlib_Image scrotGrabAutoselect(void);
 static long miliToNanoSec(int);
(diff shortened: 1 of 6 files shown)


src/scrot.h
```diff
@@ -40,4 +40,5 @@ Window scrotGetWindow(Display *, Window, int, int);
 int scrotGetGeometry(Window, int *, int *, int *, int *);
 void scrotNiceClip(int *, int *, int *, int *);
 void scrotDoDelay(void);
+void scrotWaitFor(int);
```
N-R-K, line 43: `scrotSleepFor` might've been a better name, but I tried to stay consistent with `scrotWaitUntil`.

guijan: You can change `scrotWaitUntil` to `scrotSleepUntil`, I'm indifferent to the name.

src/scrot.c
```diff
@@ -345,6 +346,24 @@ void scrotDoDelay(void)
     }
 }
 
+static inline long miliToNanoSec(int ms)
+{
+    return ms * 1000L * 1000L;
+}
```
guijan, line 352: `inline` is like `register`: a non-algorithmic optimization that the compiler can trivially figure out for us, and just like `register`, GCC and Clang ignore it entirely. Please remove that.

It only makes sense in library headers due to its semantics but even there I would avoid it.

N-R-K: Initially it was in `utils.h` but at some point I moved it back to `scrot.c` since I realized the other awkward ms => ns conversions would go away after using `scrotWaitFor`.

I'll remove the `inline` - I don't like it either.

src/scrot.c
```diff
@@ -345,6 +346,24 @@ void scrotDoDelay(void)
...
+    scrotAssert(ms >= 0);
+    struct timespec tmp;
+    clock_gettime(CONTINUOUS_CLOCK, &tmp);
+    tmp.tv_sec  += ms / 1000;
+    tmp.tv_nsec += miliToNanoSec(ms % 1000);
+    bool overflow = tmp.tv_nsec >= miliToNanoSec(1000);
+    tmp.tv_sec  += overflow;
+    tmp.tv_nsec -= overflow ? miliToNanoSec(1000) : 0;
+    scrotWaitUntil(&tmp);
+}
```
guijan, line 365: This reintroduces timer drift. The function needs to keep state, it should have a struct timespec 1st parameter that holds a timestamp, then it adds the 2nd parameter to the 1st and sleeps until that moment. 

But then if you do that you'd be reinventing `clock_nanosleep()` + `timespecadd()` when we could simply use those.

N-R-K: I don't think this "reintroduces" anything. If you take a look at the 3 call sites, 2 of them aren't called inside a loop (excluding the `EINTR` loop):

```diff
-            while (nanosleep(&t, &t) < 0 && errno == EINTR);
+            scrotSleepFor(160);

-        while (nanosleep(&t, &t) < 0 && errno == EINTR);
+        scrotSleepFor(80);
```

In these 2 cases, it's *fixing* the unlikely drift in case `EINTR` occurred.

- - -

The 3rd case is called inside a loop, but it was already drifting, so nothing is "reintroduced", but one thing is fixed: now it handles `EINTR`.

```diff
-        struct timespec delay = { 0, 50 * 1000L * 1000L };
         do {
-            nanosleep(&delay, NULL);
+            scrotSleepFor(50);
             ret = XGrabKeyboard(disp, root, False, GrabModeAsync, GrabModeAsync, CurrentTime);
         } while (--attempts > 0 && ret == AlreadyGrabbed);
```

As for keeping state, I can try it, but I suspect it'll just be awkward to use (one other reason I didn't use `scrotWaitUntil` on those "fuzzy sleeps").

N-R-K: Pushed a commit, the fact that `clock_gettime` doesn't return a value but rather uses an out-param made it more awkward than it needed to be.

(Also just realized that `clock_now` is probably not a great name if `clock_*` is reserved namespace by `time.h`/POSIX)

src/scrot.c
```diff
@@ -351,17 +351,25 @@ static long miliToNanoSec(int ms)
...
-    tmp.tv_nsec -= overflow ? miliToNanoSec(1000) : 0;
-    scrotSleepUntil(&tmp);
+    struct timespec end = {
+        .tv_sec  = (ms / 1000) + start.tv_sec,
+        .tv_nsec = miliToNanoSec(ms % 1000) + start.tv_nsec,
+    };
+    bool overflow = end.tv_nsec >= miliToNanoSec(1000);
+    end.tv_sec  += overflow;
+    end.tv_nsec -= overflow ? miliToNanoSec(1000) : 0;
+    scrotSleepUntil(&end);
```
guijan, line 371: This would be the obvious code:
```c
if (end.tv_nsec >= milliToNanoSec(1000)) {
    end.tv_sec++;
    end.tv_nsec -= miliToNanoSec(1000);
}
```

guijan:
This API could probably replace `scrotWaitUntil()` at a later point, would be less lines of code with no difference in behavior.

N-R-K:
> This API could probably replace `scrotWaitUntil()` at a later point, would be less lines of code with no difference in behavior.

Hmm, hadn't thought about it but you're probably right. I'll try to do this tomorrow with a fresher mind.

src/scrot.c
```diff
@@ -335,38 +335,67 @@ void scrotDoDelay(void)
...
          * OS X but doesn't have the macro yet.
          */
-        tmp.tv_sec = time->tv_sec - tmp.tv_sec;
-        tmp.tv_nsec = time->tv_nsec - tmp.tv_nsec;
+        tmp.tv_sec  = end.tv_sec  - tmp.tv_sec;
+        tmp.tv_nsec = end.tv_nsec - tmp.tv_nsec;
         if (tmp.tv_nsec < 0) {
-            tmp.tv_sec--;
-            tmp.tv_nsec += 1000000000L;
+            --tmp.tv_sec;
```
N-R-K, line 393: The webkit style guide doesn't say anything about preference toward pre vs post increments. And Ctrl+F "++" reveals many examples with both being used.

src/scrot.c
```diff
@@ -335,38 +335,67 @@ void scrotDoDelay(void)
...
+ * the non-standard CLOCK_BOOTTIME is used instead to avoid this bug.
  */
-static void scrotWaitUntil(const struct timespec *time)
+struct timespec clockNow(void)
 {
-    /* OpenBSD and OS X lack clock_nanosleep(), so we use a trivial algorithm to
-     * correct for drift and call nanosleep() everywhere.
-     */
+    struct timespec ret;
+#if defined(__linux__)
```
N-R-K, line 358: Didn't think the `CONTINUOUS_CLOCK` needed to be "public" anymore - since you can just use `clockNow()` instead. So the `ifdef` can be done internally for `clockNow`.

(Though the main reason I added this function was to work around the awkwardness of `clock_gettime`'s outparam, e.g `scrotSleepFor(clockNow(), 64)` is much easier to use.)

guijan:
Hmm, using the interface as is to replace `scrotWaitUntil()` means that with 32-bit int, we can wait a maximum of 24 days to take the screenshot. I wonder if that's too close for comfort?
Making that into `long long` doesn't help because then it assumes `time_t` is 64-bit which isn't necessarily the case on i386 glibc.

N-R-K:
> we can wait a maximum of 24 days to take the screenshot. I wonder if that's too close for comfort?

`opt.delay` is limited to `INT_MAX`. If someone needs more, they can use `sleep 69d && scrot` or something along those line.

N-R-K:
Ah, wait. `opt.delay` is in seconds, but this function uses `ms`. Should also reduce `opt.delay` to `INT_MAX/1000` to avoid mul overflow.

guijan:
Also, with 32-bit timespec (i.e. i386 glibc), it's possible to specify an int large enough to make struct timespec's tv_sec member overflow on master, if you write a timespec version, check for overflow.

N-R-K:
> I like pushing limits high, we just did that with filenames, but now the delay limit is falling.

I think that energy is better spent on issues like #320 or feature requests like #82 that actually concerns scrot's primary purpose - to take screenshots. There's too many other low hanging fruits to be bothering with impractical things like 24 day+ delay.

(Honestly this entire sleep saga to me seems like wasted effort fixing a non-issue that would've affected no-one.)

PR #324, opened 2023-05-27
ci: add clang-tidy

N-R-K:
the list of checks as of now is fairly conservative.

one other check that might be worth enabling in the future:

* bugprone-narrowing-conversions: unfortunately it doesn't understand
  the semantics of optionsParseNum() and thus produces way too much
  noise.

### DIFF
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -477,7 +477,7 @@ Window scrotGetWindow(Display *display, Window window, int x, int y)
     return target;
 }
 
-void scrotGrabMousePointer(const Imlib_Image image, const int xOffset,
+void scrotGrabMousePointer(Imlib_Image image, const int xOffset,
     const int yOffset)
 {
     XFixesCursorImage *xcim = NULL;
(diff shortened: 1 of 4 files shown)


src/scrot.c
```diff
@@ -942,6 +942,7 @@ static Imlib_Image scrotGrabShotMonitor(void)
     if (opt.monitor >= numScreens)
         errx(EXIT_FAILURE, "monitor %d not found", opt.monitor);
 
+    scrotAssert(screens); /* silence clang-tidy */
```
N-R-K, line 945: It doesn't understand `XineramaQueryExtension` semantics. If `numScreens` is non zero, `screens` shouldn't be NULL.

src/scrot_selection.c
```diff
@@ -92,6 +92,7 @@ static void createCursors(void)
 static void freeCursors(void)
 {
     struct Selection *const sel = *selectionGet();
+    scrotAssert(sel); /* silence clang-tidy */
```
N-R-K, line 95: This one might warrant a bit more investigation. But it seemed like an FP to me.

Issue #236, opened 2023-03-18
note.c: memory leak


guijan:
These lines of code leak:
https://github.com/resurrecting-open-source-projects/scrot/blob/a1fbb65372a2a7ce49ddd7b1041ca5f2e824a183/src/note.c#L122-L123
`parseText()` allocates a string and returns it: 
https://github.com/resurrecting-open-source-projects/scrot/blob/a1fbb65372a2a7ce49ddd7b1041ca5f2e824a183/src/note.c#L288

If you pass a font file multiple times to the note feature, it will assign the return value of `parseText()` to `note->font` without `free()`ing `note->font` first. This leaks filenames as many times as font files are repeatedly specified.

I'd personally just remove the note feature entirely, but the fix is to make sure `note->font` is initialized to `NULL` and then `free()` it before calling `parseText()`.

Edit: here's an example that leaks memory:
```console
$ scrot -n "-f '/usr/share/fonts/TTF/DroidSans-Bold/40' -f '/usr/share/fonts/TTF/DroidSans-Bold/40' -x 10 -y 20 -c 255,0,0,255 -t 'Hi'"
```

N-R-K:
I was glancing at the note code (investigating if those strdups are needed or not) and I'm left even more confused and with many more questions.

@guijan Since you have been involved in scrot development for longer, I figured I'd ask you first in case you knew already, before I put in the effort:

* If we only accept a single note, why are we free-ing and allocating a new note everytime? (`scrotNoteNew()` also frees the old note structure).

https://github.com/resurrecting-open-source-projects/scrot/blob/fafced405301d1815025f1b6194aeca1d25173af/src/options.c#L591-L602

* Why is `note` allocated? The size is known at compile time. It should just be a global var. There shouldn't be any allocation/deallocation necessary at all.

https://github.com/resurrecting-open-source-projects/scrot/blob/fafced405301d1815025f1b6194aeca1d25173af/src/note.c#L105-L107

- - -

Unrelated to notes, but one more thing I've noticed was this:

https://github.com/resurrecting-open-source-projects/scrot/blob/fafced405301d1815025f1b6194aeca1d25173af/src/scrot_selection.c#L57-L61

What's the point of this function instead of just using a global var? It looks like it's trying to do some pseudo-OOP stuff but I don't see the point, there will only ever be a single `selection` in the program.

guijan:
[...]

PR #326, opened 2023-05-28
deprecate --note

N-R-K:
reasons:

* The interface is clunky to use
* Requires Imlib2 to be built with "text" support (which adds additional dependency).
* It's better done via an actual image editing software
* No one seems to be using it, since no one objected towards removing it in the discussions thread.

ref: https://github.com/resurrecting-open-source-projects/scrot/discussions/207
ref: https://github.com/resurrecting-open-source-projects/scrot/issues/236

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -506,6 +506,8 @@ void optionsParse(int argc, char *argv[])
         opt.thumbFile = optionsNameThumbnail(opt.outputFile);
 
     if (note) {
+        warnx("--note is deprecated. See: "
+            "https://github.com/resurrecting-open-source-projects/scrot/discussions/207");
         opt.note = estrdup(note); /* TODO: investigate if dup is needed */
         scrotNoteNew(opt.note);
     }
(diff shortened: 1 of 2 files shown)


src/options.c
```diff
@@ -515,7 +517,7 @@ static void showUsage(void)
 {
     fputs(/* Check that everything lines up after any changes. */
         "usage:  " PACKAGE_NAME " [-bcfhimopuvz] [-a X,Y,W,H] [-C NAME] [-D DISPLAY]\n"
-        "              [-d SEC] [-e CMD] [-k OPT] [-l STYLE] [-M NUM] [-n OPTS]\n"
+        "              [-d SEC] [-e CMD] [-k OPT] [-l STYLE] [-M NUM]\n"
         "              [-q NUM] [-s OPTS] [-t % | WxH] [[-F] FILE]\n",
```
guijan, line 521: `[-q NUM] [-s OPTS]` could be moved to the line above them, that would make it 79 characters long.

guijan: In the output, yes. Those lines are already above 80 characters in the source code, actually.

The grey vertical bar is on column 81:
2023-05-30-175737_828x189_scrot

N-R-K: Barely fits into 80chars now.

image

Also since the options now line up in the source itself, considering deleting that comment.

PR #327, opened 2023-05-29
ci: add aggressive gcc build with -fanalyzer

N-R-K:
-fanalyzer used to throw out two warnings about memory leaks, but that seems to have been fixed with cc677e660daa53b3a8c60a18e0d0456eb322727a

also enables some aggressive gcc flags that aren't enabled by default even at -O3 (e.g graphite, ipa-pta).

### DIFF
--- a/.github/workflows/full-check.yml
+++ b/.github/workflows/full-check.yml
@@ -31,6 +31,14 @@ jobs:
         make -s distclean
         c99 -o src/scrot src/*.c $(pkg-config --cflags --libs ./deps.pc)
         src/scrot -v
+    - name: gcc-aggressive-fanalyzer
+      run: |
+        gcc --version
+        gcc -std=c99 -o src/scrot src/*.c -Wall -Wextra -Wpedantic -Werror \
+            -fanalyzer -O3 -flto -fuse-linker-plugin -fno-common \
+            -fgraphite-identity -floop-nest-optimize -fipa-pta \
+            $(pkg-config --cflags --libs ./deps.pc libbsd-overlay)
+        src/scrot -v
     - name: cppcheck
       run: |
         cppcheck --version


.github/workflows/full-check.yml
```diff
@@ -31,6 +31,14 @@ jobs:
         make -s distclean
         c99 -o src/scrot src/*.c $(pkg-config --cflags --libs ./deps.pc)
         src/scrot -v
+    - name: gcc-aggressive-fanalyzer
```
N-R-K, line 34: Wanted to do this on alpine-latest since ubuntu typically lags behind on updates (currently it's using gcc 11 when 13 is out).

But the cirrus ci yaml format is a bit different and it didn't work, I'll need to look into their docs and figure it out (probably need the `matrix` thing).

guijan: [Article on -fipa-pta](https://kristerw.blogspot.com/2017/06/fipa-pta.html). Interesting. Seems like a safe optimization as far as not backfiring and making the code bigger/slower instead goes, like `-flto` and unlike `graphite`. Might be worth researching whether we should stick it into our own default flags, the man page says it has performance issues on "large compilation units," but scrot has been sitting at ~2000LoC (give or take a few hundred) for many releases, so I assume we're not one of those.

Also, that reminds me, one optimization we can put into the code is using `restrict`. We pass and receive pointers from libraries and then use them ourselves, so we probably have many pointer dereferences that the compiler is forced to assume alias.

The GCC 11.2.0 man page says `-fno-common` is default, remove that.

N-R-K: > The GCC 11.2.0 man page says -fno-common is default, remove that.

I'm aware. But I'm also aware that when gcc switched to fno-common being the default, a lot of code broke.

It's not uncommon for distros to patch the defaults, so I'd rather keep it there. It doesn't hurt, at worse it's redundant.

guijan: Could make one AX_APPEND() call per flag.

N-R-K: > Could make one `AX_APPEND()` call per flag.

Yeah, that's roughly what I was suggesting. Feel free to open the PR.

guijan: GCC on OpenBSD doesn't have LTO, and Clang doesn't have `-fipa-pta`, Here's some preliminary `-fipa-pta` testing on OpenBSD 7.3 amd64's GCC 11.2.0 and commit e0589e4d74e738a1f6aca3ee97e445d5cea295e0 with different `-O` settings:
2023-05-30-182622_750x693_scrot
`-fipa-pta` makes consistently larger binaries, however, in the worst case scenario, they're only 0.36% bigger.

Benchmarking scrot performance is harder, we would really be measuring imlib, libpng, etc instead, perhaps it would make more sense to look up `-fipa-pta` benchmarks for other programs and see if there is a consistent performance gain over 0.36%, assuming speed and binary size are equally important.

Also, I saw no significant difference in GCC speed and memory usage with and without `-fipa-pta`:
2023-05-30-191102_657x737_scrot

N-R-K: > GCC on OpenBSD doesn't have LTO

You can use the "old school" LTO technique - that is, build everything as a single TU with `-fwhole-program`:

```console
$ cat scrot_jumbo.c
#include "src/note.c"
#include "src/options.c"
#include "src/scrot.c"
#include "src/scrot_selection.c"
#include "src/selection_classic.c"
#include "src/selection_edge.c"
#include "src/util.c"
```

On my tests, it seems to make absolutely zero difference:

```console
$ gcc --version
gcc (Gentoo 12.1.1_p20220625 p8) 12.1.1 20220625
$ time gcc -o scrot scrot_jumbo.c -fwhole-program -O3 -fipa-pta $(pkg-config --cflags --libs ./deps.pc) && wc -c scrot 
gcc -o scrot scrot_jumbo.c -fwhole-program -O3 -fipa-pta   0.36s user 0.01s system 99% cpu 0.372 total
61656 scrot
$ time gcc -o scrot scrot_jumbo.c -fwhole-program -O3  $(pkg-config --cflags --libs ./deps.pc) && wc -c scrot         
gcc -o scrot scrot_jumbo.c -fwhole-program -O3   0.34s user 0.01s system 99% cpu 0.352 total
61656 scrot
$ time gcc -o scrot scrot_jumbo.c -fwhole-program -O2 -fipa-pta $(pkg-config --cflags --libs ./deps.pc) && wc -c scrot
gcc -o scrot scrot_jumbo.c -fwhole-program -O2 -fipa-pta   0.29s user 0.02s system 99% cpu 0.309 total
57824 scrot
$ time gcc -o scrot scrot_jumbo.c -fwhole-program -O2  $(pkg-config --cflags --libs ./deps.pc) && wc -c scrot         
gcc -o scrot scrot_jumbo.c -fwhole-program -O2   0.32s user 0.01s system 99% cpu 0.327 total
57824 scrot
$ time gcc -o scrot scrot_jumbo.c -fwhole-program -Os -fipa-pta $(pkg-config --cflags --libs ./deps.pc) && wc -c scrot
gcc -o scrot scrot_jumbo.c -fwhole-program -Os -fipa-pta   0.28s user 0.02s system 99% cpu 0.292 total
49680 scrot
$ time gcc -o scrot scrot_jumbo.c -fwhole-program -Os  $(pkg-config --cflags --libs ./deps.pc) && wc -c scrot         
gcc -o scrot scrot_jumbo.c -fwhole-program -Os   0.27s user 0.02s system 99% cpu 0.287 total
49680 scrot
```

And I see no difference even with "real" LTO:

```console
$ time gcc -o scrot src/*.c -O3 -flto -fipa-pta $(pkg-config --cflags --libs ./deps.pc) && wc -c scrot 
gcc -o scrot src/*.c -O3 -flto -fipa-pta   0.47s user 0.04s system 99% cpu 0.505 total
61688 scrot
$ time gcc -o scrot src/*.c -O3 -flto $(pkg-config --cflags --libs ./deps.pc) && wc -c scrot         
gcc -o scrot src/*.c -O3 -flto $(pkg-config --cflags --libs ./deps.pc)  0.46s user 0.03s system 99% cpu 0.490 total
61688 scrot
```

I'm suspecting OpenBSD's gcc might be doing something weird. Do you get the same hash if you compile with the same settings twice? Here's the result on mine:

```console
$ gcc -o scrot src/*.c -O3 -flto $(pkg-config --cflags --libs ./deps.pc) && sha512sum scrot 
0c9267d62f3d276f19256346dba0534ab913f5dd66edce6f869f5a719e045e9a770d214984c6f1a31fd12a72bc368ecbcb85cb1f3233bcc76e9fd322d83faa5d  scrot
$ gcc -o scrot src/*.c -O3 -flto $(pkg-config --cflags --libs ./deps.pc) && sha512sum scrot
0c9267d62f3d276f19256346dba0534ab913f5dd66edce6f869f5a719e045e9a770d214984c6f1a31fd12a72bc368ecbcb85cb1f3233bcc76e9fd322d83faa5d  scrot
$ gcc -o scrot src/*.c -O3 -flto $(pkg-config --cflags --libs ./deps.pc) && sha512sum scrot
0c9267d62f3d276f19256346dba0534ab913f5dd66edce6f869f5a719e045e9a770d214984c6f1a31fd12a72bc368ecbcb85cb1f3233bcc76e9fd322d83faa5d  scrot
```

N-R-K:
Feel free to add `-fipa-pta` if you wish, now that #328 is merged. @guijan

PR #329, opened 2023-05-30
use strict C99

guijan:
Commit d149e2ed6aadc4e39b2f4b7ab9bc9fb5a4551c9b migrated us to GNU99 by using `#pragma once`, GCC and Clang aren't strict C99 compilers even in `-std=c99` mode and let this slip. Return to C99.


N-R-K:
The webkit style [recommends pragma once](https://webkit.org/code-style-guidelines/#header-guards) over header guards. And while in theory header guards are more portable, I don't think I'm aware of a single non-toy compiler that doesn't support it.

guijan:
[cparser](https://github.com/libfirm/cparser) and [pcc](http://pcc.ludd.ltu.se/) don't support it.

src/note.h
```diff
@@ -26,10 +26,13 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 */
 
-#pragma once
+#if !defined(H_NOTE)
```
N-R-K, line 29: I'd prefer `#ifndef`, it lines up with `#define` and is more idiomatic.

```suggestion
#ifndef H_NOTE
```

N-R-K (changes requested):
> Neither does MSVC, but it'll be a while before we build there.

From what I've seen, msvc's C support is laughably lackluster. So I wouldn't want to make any changes based on that.

Wikipedia [shows](https://en.wikipedia.org/wiki/C99#Implementations) cparser to have full C99 support. So that's a more convincing case. (pcc seems to be dead, last release was 2014).

guijan:
cproc is on Alpine Edge, we can add it: https://repology.org/project/cproc/versions

PCC is well packaged and gets patched by maintainers (a lot like scrot in the past), I guess Unix enthusiasts will keep it alive: https://repology.org/project/pcc/versions

cparser is not well packaged, but it's on Void Linux: https://repology.org/project/cparser/versions
But then the Void docker image is 2 years old: https://hub.docker.com/u/voidlinux. Might be abandoned? Although the repository with the dockerfiles is maintained? https://github.com/void-linux/void-docker. I don't really understand.

N-R-K (approved):

guijan:
I found a bug in PCC. It fails to compile this program:
```c
#include <stdio.h>

static void printhello(char *[static 1]);

int
main(void)
{
	char *p = "Hello, world!";
	char *hello[] = {p};
	printhello(hello);
}

static void
printhello(char *hello[static 1])
{
	puts(*hello);
}
```
PCC can't handle `static` inside the [ and ] of a declaration of a function parameter with an array type, but only if the function declaration is not a definition of a function.

Probably won't get fixed, but I reported it to their mailing list anyway. I guess this compiler won't make it to our CI for now.

guijan:
PCC fixed the bug: https://marc.info/?l=pcc-list&m=168560048728554&w=2
Also, it turns out PCC publishes tarballs regularly: http://pcc.ludd.ltu.se/ftp/pub/pcc/
So maybe an OS will update to a recent one with the bugfix in the coming months.

PR #330, opened 2023-05-30
add the TCC compiler to CI

guijan:
It uses the GNU preprocessor and linker, but it will help us with not depending on GCC too much.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -51,7 +51,7 @@ task:
       - CC: gcc
   install_script:
     - apk add autoconf autoconf-archive automake binutils gzip make pkgconfig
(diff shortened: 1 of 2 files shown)


N-R-K:
Also in terms of smaller C compilers, it might be worthwhile to add `cproc` to the CI. Although it's a C11 compiler, it *should be* capable of compiling the subset of C99 we're using (especially the fact that we're not using VLAs).

Which also reminds me, we should add `-Wvla` into one of our CI runs.

N-R-K (approved):
If this old tcc version starts bugging out, we can switch over to [alpine-testing](https://pkgs.alpinelinux.org/packages?name=tcc&branch=edge&repo=&arch=&maintainer=) which has a 2022 snapshot.

PR #331, opened 2023-05-30
src/Makefile.am: don't duplicate {CPP,C}FLAGS

guijan:
Automake always uses {CPP,C}FLAGS to compile programs. We've been erroneously duplicating them into scrot_{C,CPP}FLAGS (per-target flags).

Ref: https://www.gnu.org/software/automake/manual/html_node/Flag-Variables-Ordering.html

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -5,7 +5,7 @@
 # Copyright 2020-2021 Daniel T. Borelli <danieltborelli@gmail.com>
 # Copyright 2020      Jeroen Roovers <jer@gentoo.org>
 # Copyright 2021      Christopher R. Nelson <christopher.nelson@languidnights.com>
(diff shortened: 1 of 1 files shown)


N-R-K:
Were you able to figure out why `-O3 -Wall -Wextra -Wpedantic` is passed at linking stage even when `-flto` was disabled?

guijan:
Haven't been able to reproduce it on OpenBSD.

N-R-K:
It puts `$(CFLAGS)` in LINK. No clue why.

```console
[scrot master]~> { ./autogen.sh && ./configure; } >/dev/null
[scrot master]~> grep -F 'CFLAGS' src/Makefile 
CFLAGS =  -O3 -Wall -Wextra -Wpedantic
LINK = $(CCLD) $(AM_CFLAGS) $(CFLAGS) $(AM_LDFLAGS) $(LDFLAGS) -o $@
```

PR #332, opened 2023-05-31
ci: switch to alpine-latest for gcc -fanalyzer

N-R-K:
ubuntu lags behind too much and gcc's -fanalyzer is better on the newers versions.

ref: https://developers.redhat.com/articles/2023/05/31/improvements-static-analysis-gcc-13-compiler#


N-R-K:
Okay, so removing the backslashes fixed the yaml issue but now we're picking up warnings due to libbsd...

```
In file included from /usr/include/bsd/sys/cdefs.h:47,
                 from /usr/include/bsd/err.h:29,
                 from src/note.c:30:
/usr/include/sys/cdefs.h:4:2: error: #warning usage of non-standard #include <sys/cdefs.h> is deprecated [-Werror=cpp]
    4 | #warning usage of non-standard #include <sys/cdefs.h> is deprecated
      |  ^~~~~~~
```

Consider either going with `-Wno-cpp` or just going to standalone sys/queue and skipping libbsd for the gcc-fanalyzer run.

Issue #283, opened 2023-05-03
Enable `-Werror` in CI


N-R-K:
The CI runs should have `-Werror` enabled so that warnings don't get overlooked.

PR #333, opened 2023-06-01
ci: add -Werror

N-R-K:
this adds support for EXTRA_CFLAGS, which seems to be a common convention to append additional cflags. e.g neomutt uses it, linux kernel used to use it.

Fixes: https://github.com/resurrecting-open-source-projects/scrot/issues/283

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -24,7 +24,7 @@ common_template: &common_script
   name: ${OS}-${CC}
   build_script:
     - ./autogen.sh
-    - ./configure
+    - ./configure SCROT_PRIVATE_FLAGS="-Werror -Wno-error=cpp -Wno-error=pedantic"
     - make distcheck
   test_script:
     - make
--- a/.github/workflows/full-check.yml
+++ b/.github/workflows/full-check.yml
@@ -20,7 +20,7 @@ jobs:
     - name: distcheck
       run: |
         ./autogen.sh
-        ./configure
+        ./configure SCROT_PRIVATE_FLAGS="-Werror -Wno-error=cpp -Wno-error=pedantic"
         make distcheck
     - name: run_program
       run: |
@@ -28,7 +28,7 @@ jobs:
         src/scrot -v
     - name: tcc
       run: |
-        ./configure CC=tcc
+        ./configure SCROT_PRIVATE_FLAGS="-Werror -Wno-error=cpp -Wno-error=pedantic" CC=tcc
         make clean all
         src/scrot -v
     - name: bare build
@@ -68,10 +68,12 @@ jobs:
                   libXcomposite-devel libXext-devel libXfixes-devel libXinerama-devel \
                   make
     - name: distcheck
+      # NOTE: cygwin's imlib2 version is too old, and it's imlib_apply_filter
+      # lacks `const` qualifier. so use `-Wno-error=discarded-qualifiers`.
       run: |
         cd ${GITHUB_WORKSPACE}
         ./autogen.sh
-        ./configure
+        ./configure SCROT_PRIVATE_FLAGS="-Werror -Wno-error=cpp -Wno-error=pedantic -Wno-error=discarded-qualifiers"
         make distcheck
     - name: run_program
       run: |
--- a/configure.ac
+++ b/configure.ac
@@ -35,6 +35,13 @@ AS_IF([test "x$orig_CFLAGS" = "x"], [
             ])
         ]
     )
+    # SCROT_PRIVATE_FLAGS are mainly used by the CI to append additional flags.
+    # append them unconditionally, the CI shouldn't try to add flags the
+    # compiler doesn't support.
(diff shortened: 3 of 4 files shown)


guijan:
In Automake, CFLAGS are user compiler flags, AM_CFLAGS are project compiler flags, and foo_CFLAGS are the `foo` target's CFLAGS. We'd have 2 levels of user CFLAGS that way, just use CFLAGS. Also, `tcc` supports the C11 _Noreturn, that's a libc bug.

N-R-K:
> just use CFLAGS

The defaults of `configure.ac` will need to be needlessly duplicated that way.

> Also, `tcc` supports the C11 `_Noreturn`, that's a libc bug.

I remember [another case](https://codeberg.org/nsxiv/nsxiv/src/commit/c03ec39437b473526080f496d6c8564e98bea1d7/commands.c#L58-L59) where tcc emitted warning due to not understanding `exit()` - even though `glibc` marks it with `__attribute__((__noreturn__))`.

Funnily enough, it seems to understand `__attribute__((__noreturn__))` when manually put just fine.

```console
[/tmp]~> cat test.c
#ifdef PANIC
        extern void panic(int) __attribute__((__noreturn__));
#else
        #include <stdlib.h>
#endif

int f(_Bool x)
{
        if (x)
                return 5;
#ifdef PANIC
        panic(1);
#else
        exit(1);
#endif
}
[/tmp]~> tcc -Wall -std=c99 -c test.c -o /dev/null
test.c:16: warning: function might return no value: 'f'
[/tmp]~> tcc -Wall -std=c99 -c test.c -o /dev/null -DPANIC
```

But if I `#include <stdlib.h>` (even when using the `panic` function), then starts to warn again. Looks like something about `stdlib.h` is tripping tcc up.

guijan:
EXTRA_CFLAGS would give people another interface to fiddle with, better not.

We can try reporting the `__attribute__((__noreturn__))` or `_Noreturn` issue to glibc and see if they fix it. TCC even `#define`s \_\_GNUC__ to 4, that's why it works with OpenBSD's _err.h_.

N-R-K:
> We can try reporting the `__attribute__((__noreturn__))` or `_Noreturn` issue to glibc and see if they fix it.

I plan on sending a patch sometime this week.

> TCC even #defines __GNUC__ to 4, that's why it works with OpenBSD's err.h.

Not my my system (and an nsxiv contributor running KISS linux had the warning as well IIRC). Perhaps it's due to some OpenBSD patch?

- - -

But anyways, back on topic:

> EXTRA_CFLAGS would give people another interface to fiddle with, better not.

I went with it since it's more conventional, not because I wanted to give another env var to the user (though I don't mind giving it).

If you like, I can change it to `SCROT_PRIVATE_CFLAGS_APPEND` or something like that. I'd really like to avoid duplicating the defaults in multiple places.

guijan:
The OpenBSD `tcc` doesn't have any patches. It does, however, compile off a git commit: https://repo.or.cz/tinycc.git/commit/29ae3ed4d5b83eec43598d6cd7949bccb41c8083. Maybe that was added between whatever version your OS has and the one from OpenBSD 7.3.

Maybe the configure.ac's logic can spell out "If no CFLAGS were specified, use our defaults, and append $SCROT_PRIVATE_CFLAGS"?

N-R-K:
Done.

Also noticed that the header check is utterly broken. Here's the relavant part from `config.log`:

```
configure:5463: checking for sys/queue.h
configure:5463: gcc -c  -flto -O3 -Wall -Wextra -Wpedantic -Werror  conftest.c >&5
conftest.c:19: error: ISO C forbids an empty translation unit [-Werror=pedantic]
cc1: all warnings being treated as errors
configure:5463: $? = 1
configure: failed program was:
| /* confdefs.h */
| #define PACKAGE_NAME "scrot"
| #define PACKAGE_TARNAME "scrot"
| #define PACKAGE_VERSION "1.9"
| #define PACKAGE_STRING "scrot 1.9"
| #define PACKAGE_BUGREPORT "https://github.com/resurrecting-open-source-projects/scrot/issues"
| #define PACKAGE_URL "https://github.com/resurrecting-open-source-projects/scrot"
| #define PACKAGE "scrot"
| #define VERSION "1.9"
| #define HAVE_ERR 1
| #define HAVE_ERRX 1
| #define HAVE_WARN 1
| #define HAVE_WARNX 1
| /* end confdefs.h.  */
| 
| 
| #include <sys/queue.h>
configure:5463: result: no
```

I'll leave this up to you to fix.

N-R-K:
I've taken the easy way out by adding `-Wno-error=cpp -Wno-error=pedantic`.

Unfortunate all pedantic errors had to be disabled, but it's still an improvement over the existing state where *all* warnings go unnoticed. So I'll merge this in 1~2 days.

If you come up with a valid fix during that time, then attach the `.patch` here, I'll apply it with `git am` on this branch.

PR #336, opened 2023-06-03
[WIP] Autogenerate shell completions

N-R-K:
Recently someone mentioned [compleat](https://github.com/mbrubeck/compleat) and this made me realize we already have (almost) all the necessary information to generate shell completion in our `lopts` array.

So that's what this does. This is very much an early draft/proof of concept - so I'd like some feedback on the overall structure/direction of this (don't pay too much attention to small details at the moment).

Shells like zsh supports having a short description which can be helpful to the user, so I've added a `DESC` macro. But it's optional, so I can drop it if it's too invasive.

To test out the zsh completion yourself, run the following (or equivalent) and then reload your shell:

```console
$ tcc -run compautogen.c | doas tee /usr/local/share/zsh/site-functions/_scrot
```

Screenshot:

image

- - -

UPDATE: I think the generator is at a state where it is good enough.

However a couple other things need to be done before the PR can be merged:

- [ ] Add a autoconf/make target (or script like the `create-man.sh`) to generate the output completion files
- [ ] Make sure the generated completion files are picked up by `make dist` (I'm thinking we shouldn't commit the generated files into git)
- [ ] Add the completions to `install` target as well

The last one might need more research since the "proper directory" to install completion files might vary distro to distro.

### DIFF
--- a/etc/compautogen.c
+++ b/etc/compautogen.c
@@ -1,0 +1,81 @@
+#include <limits.h>
+#include <stdio.h>
+#include <stdbool.h>
+#include <string.h>
+
+enum { no_argument, required_argument, optional_argument, };
+struct option {
+    const char *name;
+    int has_arg;
+    int *flag;
+    int val;
+    const char *desc;
+};
+
+#define DESC(X) X
+#include "../src/opts.h"
+enum { lopts_len = (sizeof(lopts) / sizeof(lopts[0])) - 1 };
+
+static bool isDeprecated(const struct option *o)
+{
+    return !strcmp(o->name, "note") || !strcmp(o->name, "script") ||
+        !strcmp(o->name, "focussed"); /* not deprecated, but no point in a double entry */
+}
+
+static int zsh(FILE *f)
+{
+    fputs("#compdef scrot\n\n", f);
+    fputs("_arguments \\\n", f);
+    for (const struct option *o = lopts, *end = o + lopts_len; o < end; ++o) {
+        if (isDeprecated(o))
+            continue;
+
+        fprintf(f, "\t'--%s", o->name);
+        if (o->desc)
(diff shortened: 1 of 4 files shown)


compautogen.c
```diff
@@ -0,0 +1,56 @@
...
+        short_opts[n++] = o->val;
+        if (o->has_arg == required_argument)
+            short_opts[n++] = ':';
+        else if (o->has_arg == optional_argument)
+            short_opts[n++] = ':', short_opts[n++] = ':';
+    }
+    short_opts[n++] = '\0';
+    puts(short_opts);
+    return 0;
+#endif
```
N-R-K, line 35: Was a bit annoying having to adjust both `stropts` and `lopts` when adding an option so was playing around to see the viability of generating the short opts string from the lopts array.

Turned out to be bigger than I wanted it to be, so I'll probably drop this idea.

N-R-K:
I think the generator is at a state where it good enough. However a couple other things need to be done before the PR can be merged:

- [ ] Add a autoconf/make target (or script like the `create-man.sh`) to generate the output completion files
- [ ] Make sure the generated completion files are picked up by `make dist` (I'm thinking we shouldn't check in the generated files into git)
- [ ] Add the completions to `install` target as well

The last one might need more research since the "proper directory" to install completion files might vary distro to distro.

N-R-K:
I was looking at `mpv` recently and the way they do auto-completion is by having a `--list-options` flag which outputs all the flags. The completion scripts then uses that output to configure itself.

I think this is probably a better approach than doing build time generation. It'll likely also attract more contributor since there are more people who can write shell scripts than people who can (or are willing to) write generators in C.

N-R-K:
Closing in favor of https://github.com/resurrecting-open-source-projects/scrot/pull/377

PR #337, opened 2023-06-10
selection_edge: map the window once

N-R-K:
This works okay in my setup but some more testing across different set up would be nice.


N-R-K:
It looked a bit weird to me that it was mapping a mapped window again and again.

But according to the manpage:

> This function has no effect if the window is already mapped.

I guess this is fine to keep it since there's no downside.

PR #338, opened 2023-06-10
fix: default extension if --format was used

N-R-K:
There's two commits, the first one dynamically grows the buffer to add the format to the extension.

The second commit makes things simpler by placing a limit on the format length, which is far greater any any existing image formats: https://en.wikipedia.org/wiki/Image_file_format

@guijan if you don't have any objection to the 2nd commit, then I'd like it keep it since it's simpler, avoids unnecessary "leak" and unnecessary failure points created by malloc.

### DIFF
--- a/man/scrot.txt
+++ b/man/scrot.txt
@@ -93,6 +93,7 @@ SPECIAL STRINGS
 
     $$   A literal '$'.
     $a   The system's hostname.
+    $F   The output file format.
     $f   The image's full path (ignored when used in the filename).
     $h   The image's height.
     $m   The thumbnail's full path (ignored when used in the filename).
--- a/src/options.c
+++ b/src/options.c
@@ -64,6 +64,7 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
     #define PACKAGE_VERSION "unknown"
 #endif
 
+static char defaultOutputFile[] = "%Y-%m-%d-%H%M%S_$wx$h_scrot.$F";
 struct ScrotOptions opt = {
     .quality = 75,
     .compression = 7,
@@ -73,7 +74,7 @@ struct ScrotOptions opt = {
     .stackDirection = HORIZONTAL,
     .monitor = -1,
     .windowId = None,
-    .outputFile = "%Y-%m-%d-%H%M%S_$wx$h_scrot.png",
+    .outputFile = defaultOutputFile,
     .lineColor = "gray",
 };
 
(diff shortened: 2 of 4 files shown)


guijan:
No, let's keep limits in the OS and hardware.

This code has complicated requirements:
- If neither filename nor format were specified, use the default filename and format.
- If both the filename and the format were specified, use them.
- If the filename was specified and the format wasn't, infer the format from the filename. Additionally, if the filename the user gave us has no extension ("file") or the null extension ("file."), output the default format.
- If the filename wasn't specified and the format was, modify the default filename to use the format as the extension and use the modified filename with the specified format

Implementing this properly will require a little more redesigning than this PR does. 

You could have a 2-bit number, one bit is set when a custom output file is specified, and the other is set when a custom format is specified. Then you do a 4-way switch on it and handle each case individually. 

Also, I just noticed we could have an `imPrintf()` format specifier in the extension, that should either error or use the expanded string as the extension.

N-R-K:
> Then you do a 4-way switch on it and handle each case individually and in the same place.

I think all the 4 cases are handled, although not in the same place. Doing them in the same place is not that bad of an idea though. I'll look into that.

> let's keep limits in the OS and hardware.

There's no OS or hardware limit here. The limit is how many formats imlib2 supports. I don't see any reason to complicate things trying to pretend like we don't know the limit when we clearly know ahead of time the list of supported formats by imlib2.

N-R-K:
> Doing them in the same place is not that bad of an idea though. I'll look into that.

Done. Tested with all the following commands and they all produce expected results:

```sh
./scrot -e 'file --mime-type $f'
./scrot -e 'file --mime-type $f' file.webp
./scrot -e 'file --mime-type $f' file
./scrot -e 'file --mime-type $f' file.webp --format png
./scrot -e 'file --mime-type $f' --format ppm
./scrot --format webp - > out && file --mime-type out
./scrot -e 'file --mime-type $f $m' --format ppm --thumb 10
```

N-R-K:
> Also, I just noticed we could have an `imPrintf()` format specifier in the ~extension~ argument to `--format`, that should either error or use the expanded string.

This works for me. Updated the PR with commit: https://github.com/resurrecting-open-source-projects/scrot/pull/338/commits/9663ff73889f2603eb78d6fb5b02afdec871de40.

One edge case that works a bit weird now: `scrot 'file.$F'` in here `$F` will be detected as the extension and since no `--format` is set `$F` will become the format. Something like `scrot --format ppm 'file.$F'` works fine however.

daltomi:
I was testing with: `src/scrot -e 'file --mime-type $f' file`
```bash
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7c7d1ba in ?? () from /usr/lib/libc.so.6
(gdb) bt
#0  0x00007ffff7c7d1ba in ?? () from /usr/lib/libc.so.6
#1  0x00007ffff7c71bf2 in strdup () from /usr/lib/libc.so.6
#2  0x00007ffff7defb79 in imlib_image_set_format () from /usr/lib/libImlib2.so.1
#3  0x0000555555558735 in main (argc=<optimized out>, argv=<optimized out>) at src/scrot.c:156
(gdb)
```
I have seen very quickly, this condition never happens:
```c
if (!ext || opt.outputFile == defaultOutputFile)
 ```
So it happens:
```c
       else
            opt.format = ext+1; // ext = 0x0
```

N-R-K:
`scrotHaveFileExtension` doesn't set the pointer to NULL in case it didn't find an extension.

This is why in the [old commit](https://github.com/resurrecting-open-source-projects/scrot/commit/07ce507130471921f613c82e959b38c2620ae920) I had `ext` initialized to NULL. But looks like I accidentally removed it while refactoring.

guijan:
The current state of the PR looks good, but I haven't tested it.

PR #339, opened 2023-06-11
selection: remove unnecessary allocations and indirections

N-R-K:
This patch series removes some unnecessary allocations and indirections from the selection code. Everything should still continue working as usual, no change in functionality is intended.

The PR may also be better viewed as [individual commits](https://github.com/resurrecting-open-source-projects/scrot/pull/339/commits), the commit messages contain a bit more details and rationale.

### DIFF
--- a/src/selection_classic.c
+++ b/src/selection_classic.c
@@ -50,7 +50,7 @@ struct SelectionClassic {
 
 void selectionClassicCreate(void)
 {
-    struct Selection *const sel = *selectionGet();
+    struct Selection *const sel = &selection;
 
     sel->classic = ecalloc(1, sizeof(*sel->classic));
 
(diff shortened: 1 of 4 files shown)


src/selection_classic.c
```diff
@@ -86,16 +78,16 @@ void selectionClassicCreate(void)
...
         sel->rect.h);
     XFlush(disp);
 }
 
 void selectionClassicMotionDraw(int x0, int y0, int x1, int y1)
 {
-    const struct Selection *const sel = *selectionGet();
+    const struct Selection *const sel = &selection;
 
     if (sel->rect.w)
```
daltomi, line 92: This line controls the clearing of the selection.
In order for it to work correctly, you must set the values of `selection.rect` to zero (this was already done by calloc) because otherwise the following happens: `scrot -shide`
2023-06-11-040304_222x155_scrot
I would recommend to set it to zero in the function: `scrotSelectionCreate`

daltomi (changes requested):

src/selection_classic.c
N-R-K: Thanks, fixed now.

daltomi (approved):

PR #340, opened 2023-06-11
automatically detect best line mode

N-R-K:
if no compositor is running, use the "edge" mode since it has less issues compared to the classic mode.

ref: https://github.com/resurrecting-open-source-projects/scrot/issues/284


guijan (approved):
Maybe one day we can have only the "auto" mode.

PR #347, opened 2023-06-18
replace dprintf() calls to fix the SmartOS port

### LINKED ISSUE #346: build failure on SmartOS and other Solaris-based systems due to `dprintf`
SmartOS is a fork of Illumos (itself a fork of Solaris). pkgsrc's automated testing shows 1.10 doesn't compile on SmartOS because it doesn't have a `dprintf()`:
https://us-central.manta.mnx.io/pkgsrc/public/reports/upstream-trunk/20230616.2249//scrot-1.10/build.log
https://bulktracker.appspot.com/pkg/ag1zfmJ1bGt0cmFja2VyciILEgVidWlsZBiAgICjiOXCCwwLEgNwa2cYgICAs7S--gkM

Illumos has online documentation, there's no `dprintf()` there, so we're broken on Illumos also: https://illumos.org/man/3C/all
It's likely the same is true for Solaris, but it seems their web conversion of the man pages is broken: https://docs.oracle.com/cd/E19253-01/817-0550/euqfl/index.html.

guijan:
I don't have access to a SmartOS system, so I haven't tested it on SmartOS. I did test it on OpenBSD.

Fixes #346.

### DIFF
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -332,12 +332,17 @@ void scrotDoDelay(void)
     if (!opt.delay)
         return;
     if (opt.countdown) {
-        dprintf(STDERR_FILENO, "Taking shot in ");
+        fputs("Taking shot in ", stderr);
         for (int i = opt.delay; i > 0; i--) {
-            dprintf(STDERR_FILENO, "%d.. ", i);
+            /* Illumos doesn't have dprintf():
+             * https://www.illumos.org/issues/1609
+             */
+            fprintf(stderr, "%d.. ", i);
+            fflush(stderr);
             opt.delayStart = scrotSleepFor(opt.delayStart, 1000);
         }
-        dprintf(STDERR_FILENO, "0.\n");
+        fprintf(stderr, "0.\n");
+        fflush(stderr);
     } else {
         scrotSleepFor(opt.delayStart, opt.delay * 1000);
     }


N-R-K:
I'm ok merging this. But `dprintf` has been standard since POSIX-2008, which is ~15 years old now. We should at the very least file a bug against Illumos about this and link the bug in the commit message.

Adding a comment (possibly with the Illumos bug link) is probably also a good idea to avoid someone accidentally using dprintf again (since we don't run Illumos CI, it won't be caught easily).

guijan:
They have an 11 year old report of this: https://www.illumos.org/issues/1609

src/scrot.c
```diff
@@ -332,12 +332,18 @@ void scrotDoDelay(void)
     if (!opt.delay)
         return;
     if (opt.countdown) {
-        dprintf(STDERR_FILENO, "Taking shot in ");
+
```
N-R-K, line 335: Any reason for the blank line?

N-R-K (changes requested):

N-R-K:
Also please add `Fixes #346` into the commit message as well.

N-R-K (approved):
I thought the fflush could be dropped since stderr is unbuffered, but after looking at the standard, [it says](https://port70.net/~nsz/c/c99/n1256.html#7.19.3p7) it's "not fully buffered" - which leaves room for it to be line-buffered still.

I guess we can set to to unbuffered at startup ourself, but meh, think we can leave this be.

PR #349, opened 2023-06-22
set blending option once, before the loop

N-R-K:
instead of setting them everytime in the loop, it's sufficient to set it just once before entering the loop.

also remove imlib_context_set_angle() since that is a text related setting and has no effect on imlib_blend_image_onto_image().

### DIFF
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -1006,6 +1006,10 @@ static Imlib_Image stalkImageConcat(ScrotList *images, const enum Direction dir)
     imlib_context_set_color(0, 0, 0, 255);
     imlib_image_fill_rectangle(0, 0, w, h);
 
+    imlib_context_set_anti_alias(0);
+    imlib_context_set_dither(1);
+    imlib_context_set_blend(0);
+
     image = firstScrotList(images);
     while (image) {
         im = (Imlib_Image) image->data;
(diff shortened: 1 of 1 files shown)


N-R-K:
> remove imlib_context_set_angle() since that is a text related setting

I noticed this when trying to link against imlib2 built with `--disable-text` and saw the error:

```
src/scrot.c:1016: undefined reference to `imlib_context_set_angle'
```

This was on a local branch with scrot's `--notes` feature removed.

- - -

Everything should still work fine, I think. But I do not use a compositor so I did not test. If someone can test and confirm everything to be working, then I can merge this.

N-R-K:
I just removed this check and tried `scrot -k` and everything worked perfectly fine without running any compositor.

https://github.com/resurrecting-open-source-projects/scrot/blob/5c2d6afe801ee36d0744e6e89c2cf4f8156dcff1/src/scrot.c#L832-L836

I don't think a compositor is needed for `-k` since we redirect all the sub-windows manually here:

https://github.com/resurrecting-open-source-projects/scrot/blob/5c2d6afe801ee36d0744e6e89c2cf4f8156dcff1/src/scrot.c#L866

- - -

Here's the resulting screenshot:

scrot

N-R-K:
Tested this PR with that check removed and everything seems to be working fine.

- - -

Also opened a new PR for removing the compositor check: #354. So let's move any discussion about that over to the new PR.

daltomi (approved):
It looks good.

PR #354, opened 2023-06-25
don't require a compositor for -k

N-R-K:
compositor doesn't seem to be needed and works fine without it. See: https://github.com/resurrecting-open-source-projects/scrot/pull/349#issuecomment-1606244433

### DIFF
--- a/man/scrot.txt
+++ b/man/scrot.txt
@@ -42,8 +42,9 @@ OPTIONS
   -h, --help                Display help and exit.
   -i, --ignorekeyboard      Don't exit for keyboard input. ESC still exits.
   -k, --stack[=OPT]         Capture stack/overlapped windows and join them. A
-                            running Composite Manager is needed. OPT it's optional
-                            join letter: v/h (vertical/horizontal). Default: h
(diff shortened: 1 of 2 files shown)


daltomi:
It doesn't work as expected. The first image (above) with CWM enabled captures the entire video area while without CWM it does not (below image).
Note: the "BSD" video window is partially hidden by the xterm window. Both videos are in play mode.

2023-06-25-184959_2023x480_scrot
2023-06-25-185009_2023x480_scrot

N-R-K:
I see, here is the result on dwm (without any compositor):

* When the video window is tiled and the xterm window is floating, it works OK.

* But if both window is floating then the xterm window gets into the video (screenshot below).

image

- - -

@daltomi How about we change this to a warning instead of error: "--stack: Compositor is not running, some windows may overlap"?

daltomi:
> @daltomi How about we change this to a warning instead of error: "--stack: Compositor is not running, some windows may overlap"?
Yes, looks good to me. Remember to update the man page too :-)

PR #356, opened 2023-06-27
Use an array instead of linked list

N-R-K:
### stalkImageConcat: use an array

there's no algorithmic reason to use a list over an array here.
an array is both simpler and avoids depending on non-standard header.

### ci: remove -Wno-error=pedantic

we no longer check for any headers. but keep the comment in configure.ac
for the future.


N-R-K:
Hmm, CI seems to be failing for macos:

```
checking for err... no
checking for errx... no
checking for warn... no
checking for warnx... no
checking for libbsd-overlay... no
```

I suspect this might be due to commit https://github.com/resurrecting-open-source-projects/scrot/pull/356/commits/81b2de6b480e0426131929e526f7159434a37696. I cannot reproduce this locally, is there anyway to inspect the `config.log` from the failed CI run?

N-R-K:
Figured out how to get the `config.log`.

```
configure:5407: checking for errx
configure:5407: clang -o conftest  -flto -O3 -Wall -Wextra -Wpedantic -Werror -Wno-error=cpp   conftest.c  >&5
conftest.c:27:11: error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]
char errx ();
          ^
           void
1 error generated.
```

Wow, this looks really bad... I think this is because newer clang started erroring out on prototypes with unspecified arguments (I believe gcc plans to do the same). So we definitely need to fix this.

I think the fix would be to tell autoconf to include `<err.h>` instead of making up a nonsensical prototype.

ping: @guijan

N-R-K:
I thought this change would've also allowed us to remove `-Wno-error=pedantic` from the CI. But I guess not, just going to revert https://github.com/resurrecting-open-source-projects/scrot/commit/81b2de6b480e0426131929e526f7159434a37696 and leave it for another time.

PR #362, opened 2023-07-03
fix: bad extension detection

N-R-K:
in a command like `scrot ~/.local/testfile` scrot will incorrectly think that `local/testfile` is the file extension and thus will try to use it as the output format.

ensure that the extension doesn't contain any slashes to avoid such issues.

### DIFF
--- a/.github/workflows/full-check.yml
+++ b/.github/workflows/full-check.yml
@@ -41,6 +41,7 @@ jobs:
       run: |
         cppcheck --version
         cppcheck --std=c99 -j$(nproc) --quiet --force --error-exitcode=1 \
+            --suppress=uninitvar \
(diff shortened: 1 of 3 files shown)


N-R-K:
The thumbnail filename logic (which doesn't use `scrotHaveFileExtension()`) also suffers from this issue.

N-R-K:
Well, it still doesn't quite fix the issue. `scrot .dotpic` will treat `dotpic` as the extension but prefixing a filename with `.` in order to hide them is common in unix-like systems.

So we'll probably want to special case leading `.` as well.

N-R-K:
I think this takes care of all the edge cases (excluding the thumbnail logic which hasn't been touched yet).

Issue #271, opened 2023-04-09
scrot exits success on error


guijan:
Scrot exits success even if the `-e` option fails:
https://github.com/resurrecting-open-source-projects/scrot/blob/75664efd19d02c0d70006e902364a877bec902b9/src/scrot.c#L572-L579

guijan:
I also just noticed that you can make your command this:
```sh
scrot -e 'return 127'
```
Scrot is going to incorrectly complain it couldn't execute the command. The POSIX shell may be defined to return 127 if it can't find the command, but nothing stops the last program in shell script from returning 127 and thus making the shell return 127.

Also, the shell will probably print an error message for us anyway, so that is redundant:
![2023-06-12-094602_496x64_scrot](https://github.com/resurrecting-open-source-projects/scrot/assets/25590950/feae4967-323d-45ac-a4f9-97537c691d65)

N-R-K:
What's the sensible thing to do if the child exits with non-zero status? Should we mirror it? But then again non-zero exit status doesn't necessarily mean an error and some programs use exit status to signal different things (e.g `test(1)`) which aren't exactly errors.

guijan:
We could mirror it & not print anything. That would fix all the issues mentioned here. But then it raises the issue of not knowing whether a non-0 return is from scrot or the command.

Or we could return 0 if scrot & the command succeed, 1 if scrot errors, 2 if the command returns non-0. Then, the exact return value of the command isn't returned, but you can tell who returned non-0. It's okay of `scrot -e` has limitations, if someone wants something more powerful he can write a shell script.

N-R-K:
Rough patch:

```diff
diff --git a/src/scrot.c b/src/scrot.c
index 3cb1dcf..9bf56be 100644
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -644,11 +644,10 @@ static void scrotExecApp(Imlib_Image image, struct tm *tm, char *filenameIM,
 {
     char *execStr = imPrintf(opt.exec, tm, filenameIM, filenameThumb, image);
     int ret = system(execStr);
-
     if (ret == -1)
-        warn("The child process could not be created");
-    else if (WIFEXITED(ret) && WEXITSTATUS(ret) == 127)
-        warnx("scrot could not execute the command: %s", execStr);
+        err(EXIT_FAILURE, "The child process could not be created");
+    else if (WIFEXITED(ret) && WEXITSTATUS(ret) != 0)
[...]

PR #370, opened 2023-09-24
don't return 0 on exec failure

N-R-K:
additionally return EX_UNAVAILABLE [^0] if the child terminated abnormally or returned non-zero.

[^0]: https://man.freebsd.org/cgi/man.cgi?sektion=3&query=sysexits

Closes: https://github.com/resurrecting-open-source-projects/scrot/issues/271

### DIFF
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -633,11 +633,10 @@ static void scrotExecApp(Imlib_Image image, struct tm *tm, char *filenameIM,
 {
     char *execStr = imPrintf(opt.exec, tm, filenameIM, filenameThumb, image);
     int ret = system(execStr);
-
     if (ret == -1)
-        warn("The child process could not be created");
-    else if (WIFEXITED(ret) && WEXITSTATUS(ret) == 127)
-        warnx("scrot could not execute the command: %s", execStr);
+        err(EXIT_FAILURE, "The child process could not be created");
+    else if (!WIFEXITED(ret) || WEXITSTATUS(ret) != 0)
+        exit(69 /* EX_UNAVAILABLE */ );
     free(execStr);
 }
 


guijan:
It can't be merged as is because of https://github.com/resurrecting-open-source-projects/scrot/issues/271#issuecomment-1734754971

src/scrot.c
```diff
@@ -644,11 +644,10 @@ static void scrotExecApp(Imlib_Image image, struct tm *tm, char *filenameIM,
...
     char *execStr = imPrintf(opt.exec, tm, filenameIM, filenameThumb, image);
     int ret = system(execStr);
-
     if (ret == -1)
-        warn("The child process could not be created");
-    else if (WIFEXITED(ret) && WEXITSTATUS(ret) == 127)
-        warnx("scrot could not execute the command: %s", execStr);
+        err(EXIT_FAILURE, "The child process could not be created");
+    else if (!WIFEXITED(ret) || WEXITSTATUS(ret) != 0)
+        exit(69 /* EX_UNAVAILABLE */ );
```
guijan, line 650: It can't be merged as is because of https://github.com/resurrecting-open-source-projects/scrot/issues/271#issuecomment-1734754971
POSIX only defines what WEXITSTATUS() does if WIFEXITED() returns true.

N-R-K: I'm a bit baffled here. If `WIFEXITED` was false then the exit status macro will not be evaluated due to the logical OR. What am I missing?

guijan: That's because I'm a bad reader.
Still, it should be:
```c
else if (WIFEXITED(ret) && WEXITSTATUS(ret) || WTERMSIG(ret) && WIFSIGNALED(ret)
```

N-R-K: >  Still, it should be:

Is there any functional difference here or just stylistic? AFAICS, the `!WIFEXITED(..)` should already catch `WIFSIGNALED(..)` case.

N-R-K: Hmm, I see. I've made the check more explicit.

> FreeBSD has a wait4() function that returns if the child hits a breakpoint in a debugger

On a sidenote: Shouldn't this be considered a bug in freebsd if system returns without the child terminating?

guijan: > > FreeBSD has a wait4() function that returns if the child hits a breakpoint in a debugger
> 
> On a sidenote: Shouldn't this be considered a bug in freebsd if system returns without the child terminating?

Yes.

guijan: Hmm, actually, the `WTERMSIG(ret)` bit should be removed because it won't be optimized out and `0` is the null signal. The null signal can't happen.

guijan: After further research and writing a test program, it appears that my interpretation of the way FreeBSD's `ptrace()` syscall works is wrong.

If a process doesn't use the `ptrace()` syscall, `system()` will never return because the child stopped. Only a process that is using the `ptrace()` API can have a `wait*()` function return `WIFSTOPPED()` even if the `WUNTRACED` flag wasn't specified. scrot doesn't use `ptrace()` and never will because that's for writing debuggers.

N-R-K: So what's the conclusion? IMO the first version was short and read nicely: "If it wasn't a normal exit or the exit code was non-zero then exit with EX_UNAVAILABLE."

guijan (approved):

PR #371, opened 2023-09-30
Correct strftime manual section

tricantivu:
Previously, scrot's manual mentioned the manual `strftime(2)`, which, as far as
I know it is incorrect. According to the Linux manpage `man-pages(7)`,  section
number `2` is for system calls and not for library functions.

I attempted searching for the manual `strftime(2)` to no avail with the command
`man -s 2 --names-only --wildcard 'strftime(2)'` and there are no lines with
the word `strftime` in the file `/usr/include/asm/unistd_64.h`, which has
system call numbers.

### DIFF
--- a/man/scrot.txt
+++ b/man/scrot.txt
@@ -88,7 +88,7 @@ OPTIONS
 SPECIAL STRINGS
   -e, -F and FILE parameters can take format specifiers that are expanded
   by scrot when encountered. There are two types of format specifier:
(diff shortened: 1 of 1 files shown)


N-R-K (approved):

N-R-K:
Thanks, merged.

tricantivu:
No problem, thanks for reviewing.

PR #372, opened 2023-10-09
autogen.sh shell script rewrite

tricantivu:
I have seen the pull request #174 by user @guijan, however, I disagree with the
code style and the argument against using `echo`.

I believe it is easier to validate the `autogen.sh` shell script command line
arguments and, albeit more readable by just expanding the first positional
parameter and ignoring the rest. Furthermore, the `autogen.sh` shell script
from #174 would consider an empty string as valid (e.g., `./autogen.sh ''`).

If the script were to become more complex, [arguments can be parsed in other ways](https://github.com/Kicksecure/sandbox-app-launcher/blob/master/usr/bin/sandbox-app-launcher#L8).

Regarding the use of the `echo` command, there should not be any significant
portability issue because all strings or arguments do not contain escape
sequences to be interpreted. When more formatting is required, `echo` can be
replaced to prevent errors because of the differences in their options. For
more information:

- https://man.openbsd.org/csh.1#echo
- https://man.openbsd.org/ksh.1#echo~2
- https://www.gnu.org/software/bash/manual/bash.html#index-echo
- https://git.kernel.org/pub/scm/utils/dash/dash.git/tree/src/bltin/printf.c#n458

While editing the `autogen.sh` shell script I noticed there was a
backslash near the end of a line with an `rm` command. Is there some line
length limit that I am not aware of?

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -29,29 +29,32 @@
 # OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
 # SUCH DAMAGE.
 
+[ "$#" -ge 1 ] && {
+    case "$1" in
+        (clean)
+            if [ -e Makefile ]; then
+                echo "I can not clean. Use 'make distclean'." >&2
+                exit 1
+            else
+                echo "Vanishing the code"
+                rm -rf aclocal.m4 autom4te.cache/ compile configure depcomp install-sh \
+                       Makefile.in missing src/config.h.in src/Makefile.in
+                exit
+            fi
+        ;;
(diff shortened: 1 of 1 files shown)


guijan:
The backslash is to avoid going over 80 columns in a line.

tricantivu:
> The backslash is to avoid going over 80 columns in a line.

Should I rebase my commits?

guijan:
> Should I rebase my commits?

Yes, break lines before they go over 80 columns.

N-R-K:
Thanks for the PR, I've went with something simpler based on #174.

PR #377, opened 2024-05-27
Add `--list-options` and shell completions

N-R-K:
This adds a `--list-options` flag which outputs a specialized TSV of all the options intended to be used in shell completion scripts. A couple TODOs:

- [x] Make `--list-options` accept an optional argument. By default it should output human readable format, but when using `--list-options=tsv` it would output the TSV.
- [ ] `make install` should install the zsh completion script
- [x] Completion script for bash
- [x] Add `--list-options` to manpage etc
- [x] Should `--help` output also include (human readable) list of options??

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -38,7 +38,7 @@ man_MANS = man/scrot.1
 
 dist_doc_DATA = AUTHORS ChangeLog CONTRIBUTING.md doc/scrot.png FAQ.md README.md TODO.md
 
-EXTRA_DIST = $(man_MANS) autogen.sh deps.pc
+EXTRA_DIST = $(man_MANS) autogen.sh deps.pc etc/
 
 SUBDIRS = src
 
--- a/README.md
+++ b/README.md
@@ -99,6 +99,9 @@ Debian users can install scrot from their package manager:
 $ sudo apt install scrot
 ```
 
+Bash and Zsh completion scripts are available in [etc/](./etc).
+Distro packagers are encouraged to install them to the appropriate directory.
+
 ## Author ##
 
 scrot was originally developed by Tom Gilbert.
--- a/TODO.md
+++ b/TODO.md
@@ -11,6 +11,17 @@ Look for any bugs reported downstream or in our github issues and fix them:
 - <https://bugzilla.redhat.com/buglist.cgi?bug_status=NEW&bug_status=ASSIGNED&bug_status=REOPENED&component=scrot&product=Fedora>
 - <https://bugs.gentoo.org/buglist.cgi?quicksearch=scrot>
 
+## Shell completion improvements
+
+- Improve the bash completion script. Currently it's quite rudimentary.
+- Filter out mutually exclusive options in zsh completion. E.g if `-b` is
+  already provided then `--border` should no longer be considered a potential
+  match.
+- Add some way to install zsh/bash completion scripts. Since the directory where
+  completion script go into can vary, we want the user to explicitly set the
+  installation directory. E.g via the configure script
+  `./configure --with-zsh-completion-path=...`.
+
 ## Switch to newer Imlib2 interfaces
 
 These will require a minimum version bump on imlib2 and so has to be done with
--- a/etc/bash-completion/scrot
+++ b/etc/bash-completion/scrot
@@ -1,0 +1,29 @@
+#!/bin/bash
+#
+# Copyright 2024 NRK <nrk@disroot.org>
+#
+# This is free and unencumbered software released into the public domain.
+# For more information, please refer to <https://unlicense.org/>
+
+_scrot_autocomp() {
+	local -a _scrot_opt_list
+	IFS=$'\n' _scrot_opt_list=( $(scrot --list-options=tsv) ) 2>/dev/null
+	local _scrot_opts=""
+	for ln in "${_scrot_opt_list[@]}"; do
+		local -a tokens
+		IFS=$'\t' tokens=( ${ln} )
+
+		local sopt="${tokens[0]}"
+		local lopt="${tokens[1]}"
+		# TODO: better support for flags that accept argument
+		# local argtype=${tokens[2]%%:*}
+		# local argdesc=${tokens[2]#*:}
+		# local desc=${tokens[3]}
+		_scrot_opts+="--${lopt} "
+		if [[ "$sopt" = [[:alnum:]]* ]]; then
+			_scrot_opts+="-${sopt} "
+		fi
+	done
+	printf "%s" "${_scrot_opts% }"
+}
+complete -W "$(_scrot_autocomp)" scrot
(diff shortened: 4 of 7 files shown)


N-R-K:
> `make install` should install the zsh completion script

Not sure how to tackle this one. Shell completion script dirs are not as well "standardized" as `/usr/local/`. For example, older zsh version doesn't recognize `/usr/local/share/zsh/site-functions` by default. On the other hand, installing into `/usr/share..` violates the assumption that `make install` files goes to `/usr/local...`.

Maybe we shouldn't try to install the completion files and just let distro packagers install it to appropriate directory for the respective distro. Or make it so that completion files will be installed only when explicitly requested via something like `./configure --with-zsh-completion-path="..."` or `make install ZSH_COMP_PATH="..."`.

N-R-K:
> Should --help output also include (human readable) list of options??

I'm thinking we could probably just remove the current usage string entirely and utilize the new `--list-options` for help string:

```console
$ ./scrot -h
Usage:  scrot [OPTIONS] [FILE]

OPTIONS
=======
-a, --autoselect <x,y,w,h>      autoselect provided region
-b, --border                    capture the window borders as well
-C, --class <NAME>              capture specified window class
-c, --count                     display a countdown for delay
-D, --display <DISPLAY>         capture specified display
-d, --delay <[b]SEC>            add delay before screenshot
-e, --exec <CMD>                execute command on saved image
-F, --file <FILE>               specify output file
-f, --freeze                    freeze the screen when -s is used
-h, --help                      display help and exit
-i, --ignorekeyboard            ignore keyboard
-k, --stack[=v|h]               capture overlapped window and join them
-l, --line <STYLE>              specify the style of the selection line
-M, --monitor <NUM>             capture monitor
-m, --multidisp                 capture all monitors
-o, --overwrite                 overwrite the output file if needed
-p, --pointer                   capture the mouse pointer as well
-q, --quality <NUM>             image quality
-s, --select[=OPTS]             interactively select a region to capture
-t, --thumb <% | WxH>           also generate a thumbnail
-u, --focused                   capture the currently focused window
-u, --focussed                  capture the currently focused window
-v, --version                   output version and exit
-w, --window <WID>              X window ID to capture
-Z, --compression <LVL>         image compression level
-z, --silent                    prevent beeping
--format <FMT>                  specify output file format
--list-options[=human|tsv]      list all options
```

This seems significantly more helpful than the current `-h` output.

N-R-K:
Any feedback on this? I'd like to merge this and cut a release sometime this week.

daltomi:
> This seems significantly more helpful than the current `-h` output

Yes I agree. Previously the format was like this but then it was decided to change it, see: b4ca02e ("use the manual's synopsis as help text (#104)", 2021-07-28)

daltomi:
In this one, as in previous commits ;-) the copyright date needs to be updated.

daltomi:
> `make install ZSH_COMP_PATH="..."`.

I like this one better, if possible, I don't know. But making a target in Makefile ala `make install-shell'  or something like that would be best.

daltomi:
It would be necessary to add the `etc/` directory, so that when doing `make dist`  it would include it.
```patch
diff --git a/Makefile.am b/Makefile.am
index 63e275e..497b212 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -40,6 +40,8 @@ dist_doc_DATA = AUTHORS ChangeLog CONTRIBUTING.md doc/scrot.png FAQ.md README.md
 
 EXTRA_DIST = $(man_MANS) autogen.sh deps.pc
 
+EXTRA_DIST += etc/
+
 SUBDIRS = src
 
 distclean-local:
```

N-R-K:
> It would be necessary to add the `etc/` directory, so that when doing `make dist` it would include it.

Good catch. Fixed.

> the copyright date needs to be updated.

I will go through all the commits since the previous version and update the copyright date in another commit.

> "use the manual's synopsis as help text (https://github.com/resurrecting-open-source-projects/scrot/pull/104)"

That was okay in the past, but currently we have long options which have no short options (e.g `--format`). Showing only short opt excludes these long only opt.

daltomi:
> Or make it so that completion files will be installed only when explicitly requested via something like `./configure --with-zsh-completion-path="..."`

Hi,
I was doing some tests and as you indicate it can be done (It is similar but with environment variable), but not with `make install-shell` as I would like :-/
I attach the patch.
[scrot.txt](https://github.com/user-attachments/files/15540177/scrot.txt)
This works as follows. First, clarify that the names of the variables/options are vague, it's just an idea.

The new options are:
```bash
 --with-zsh-confpath     install autocomplete zsh files
 --with-bash-confpath    install autocomplete bash files
```
Example:
```bash
$ ./configure --with-zsh-confpath
configure: error: the variable ZSH_CONFPATH no set
```
It expects to find a `ZSH_CONFPATH `environment variable defined. We define it like this( or whatever, export, set, etc)
```bash
$  ./configure --prefix=/usr/local --with-zsh-confpath ZSH_CONFPATH=/tmp/zsh-autocompletion
```
Then we install to test:
```bash
make DESTDIR=/tmp/scrot install
```
It is installed on: `/tmp/scrot/usr/local` and `/tmp/zsh-autocompletion/`
Note that the `--prefix` option had no impact on the installation of zsh

That is an approximation, it would be necessary to see if there is a better or more robust one because this patch does not uninstall the shell files for example.

N-R-K:
> Note that the --prefix option had no impact on the installation of zsh

I think that's probably fine, but I'm not sure.

> configure: error: the variable ZSH_CONFPATH no set

Can we make it so that the path is given as an argument? E.g:

```console
$ ./configure --with-zsh-confpath=/tmp/zsh-complete
```

I think this is more easy to use.

N-R-K:
I've merged this PR since the autocompletion itself works.

> I attach the patch.

For installing the files, you can now create your own PR. I'm not good with autoconf, so I wouldn't be able to do it anyways.

daltomi:
El Tue, 04 Jun 2024 00:06:27 -0700
NRK ***@***.***> escribió:
> I've merged this PR since the autocompletion itself works.
> 
> > I attach the patch.  
> 
> For installing the files, you can now create your own PR. I'm not good with autoconf, so I wouldn't be able to do it anyways.
> 

Ok, as soon as I can I'll make a PR and we'll continue it there.

Regards.

N-R-K:
@daltomi I was thinking it might be time for a release. Do you think we should wait to add autoconf/makefile completion installation before making a release?

I think most distros/package-managers will not have any trouble installing those files.

daltomi:
El Sun, 09 Jun 2024 11:28:23 -0700
NRK ***@***.***> escribió:
> @daltomi I was thinking it might be time for a release. Do you think we should wait to add autoconf/makefile completion installation before making a release?
> 
> I think most distros/package-managers will not have any trouble installing those files.
> 

Hi,
Yes, it's okay with me.

PR #378, opened 2024-05-29
add sleep to classic selection destroy as well

N-R-K:
selection classic is still the default for many people who are using a compositor. adding a hacky sleep similar to selection edge seems to fix it on my end.

Fixes: https://github.com/resurrecting-open-source-projects/scrot/issues/284

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -16,17 +16,6 @@ into the selection area you have created.
 Use the `--ignorekeyboard` option. This option prevents exit by
 ignoring any keyboard event, except the <kbd>ESC</kbd> key.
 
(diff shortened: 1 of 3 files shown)


N-R-K:
For testing I used this to select a video playing in firefox:

```console
$ ./scrot -s -lmode=classic,width=8,color=#ff3838
```

9/10 times the (discolored) border would end up in the screenshot. After the patch, I haven't been able to reproduce it.

(Should also note that I'm not running any compositor. And `edge` mode never seems to get into the screenshot for me, with or without the sleep.)

PR #385, opened 2024-09-14
(man page) add extra space to make -u line render as heading

k4gi:
The line in the man page for `-u, focused, --focussed` is not being rendered properly. Adding an extra space between the flags and the description in scrot.txt fixes this.

This does put it out of line with the rest of the descriptions...


N-R-K (approved):
> This does put it out of line with the rest of the descriptions...

Unfortunate. But I couldn't find anything better from playing around a bit.

Issue #179, opened 2022-01-11
Right click to abort selection


tunalad:
This might be subjective, but I think most people use the left click when using the selection option. I think it would be a lot better if we just could hit the right mouse button and cancel the operation, instead of having to hit a key on the keyboard.

I know that I can simply do that myself, but I still feel like it's a pretty good option for everyone using this tool.

daltomi:
The problem is that the user may currently be using the right mouse button to capture, either with a simple click or by selecting an area.

Another option would be, once the selection area has started, determine which button (left / right) the user used and capture the event of the other button as a cancel.
But this is only possible once the user creates a **selection area** and maintains it; at that moment the other mouse button is pressed, which is not currently being used for the selection, canceling the capture.

```patch
diff --git a/src/scrot_selection.c b/src/scrot_selection.c
index d09b3d2..bae1cf5 100644
--- a/src/scrot_selection.c
+++ b/src/scrot_selection.c
@@ -242,6 +242,7 @@ bool scrotSelectionGetUserSel(struct SelectionRect *selectionRect)
     fd_set fdSet;
     int count = 0, done = 0;
     int rx = 0, ry = 0, rw = 0, rh = 0, isButtonPressed = 0;
+    unsigned int buttonPressedId = 0;
     Window target = None;
     Status ret;
 
@@ -275,6 +276,7 @@ bool scrotSelectionGetUserSel(struct SelectionRect *selectionRect)
                 break;
             case ButtonPress:
                 isButtonPressed = 1;
+                buttonPressedId = ev.xbutton.button;
                 rx = ev.xbutton.x;
                 ry = ev.xbutton.y;
                 target = scrotGetWindow(disp, ev.xbutton.subwindow, ev.xbutton.x, ev.xbutton.y);
@@ -283,6 +285,8 @@ bool scrotSelectionGetUserSel(struct SelectionRect *selectionRect)
                 break;
             case ButtonRelease:
                 done = 1;
+                if (buttonPressedId != ev.xbutton.button)
+                    done = 2;
                 break;
             case KeyPress:
             {
```

PR #388, opened 2024-11-06
handle left and right mouse buttons only for selection

7xnl:
Left mouse (button 1) selects, right mouse (button 3) aborts. Other mouse buttons are ignored. Mirrors behavior in similar tools.

Fixes: https://github.com/resurrecting-open-source-projects/scrot/issues/179

### DIFF
--- a/src/scrot_selection.c
+++ b/src/scrot_selection.c
@@ -232,15 +232,22 @@ static bool scrotSelectionGetUserSel(struct SelectionRect *selectionRect)
                 scrotSelectionMotionDraw(rx, ry, ev.xmotion.x, ev.xmotion.y);
             break;
         case ButtonPress:
-            isButtonPressed = true;
-            rx = ev.xbutton.x;
-            ry = ev.xbutton.y;
-            target = scrotGetWindow(disp, ev.xbutton.subwindow, ev.xbutton.x, ev.xbutton.y);
-            if (target == None)
-                target = root;
+            if (ev.xbutton.button == 1) {
+                isButtonPressed = true;
+                rx = ev.xbutton.x;
+                ry = ev.xbutton.y;
+                target = scrotGetWindow(disp, ev.xbutton.subwindow, ev.xbutton.x, ev.xbutton.y);
+                if (target == None)
+                    target = root;
+            }
             break;
         case ButtonRelease:
-            done = DONE;
+            if (ev.xbutton.button == 1)
+                done = DONE;
+            else if (ev.xbutton.button == 3) {
+                warnx("Right mouse button was pressed, aborting shot");
+                done = ABORT;
+            }
             break;
         case KeyPress:
         {


src/scrot_selection.c
```diff
@@ -232,15 +232,22 @@ static bool scrotSelectionGetUserSel(struct SelectionRect *selectionRect)
...
+                if (target == None)
+                    target = root;
+            }
             break;
         case ButtonRelease:
-            done = DONE;
+            if (ev.xbutton.button == 1)
+                done = DONE;
+            else if (ev.xbutton.button == 3) {
+                warnx("Right mouse button was pressed, aborting shot");
```
N-R-K, line 248: It says "pressed" but this is operating on release, not press.

N-R-K: Looks like it doesn't generate a button press event if one of the other button is held down?

```console
[scrot master]~> git diff
diff --git a/src/scrot_selection.c b/src/scrot_selection.c
index 9483e9c..8aaf666 100644
--- a/src/scrot_selection.c
+++ b/src/scrot_selection.c
@@ -232,6 +232,7 @@ static bool scrotSelectionGetUserSel(struct SelectionRect *selectionRect)
                 scrotSelectionMotionDraw(rx, ry, ev.xmotion.x, ev.xmotion.y);
             break;
         case ButtonPress:
+            warnx("got ButtonPress (%u)", ev.xbutton.button);
             isButtonPressed = true;
             rx = ev.xbutton.x;
             ry = ev.xbutton.y;
@@ -240,6 +241,7 @@ static bool scrotSelectionGetUserSel(struct SelectionRect *selectionRect)
                 target = root;
             break;
         case ButtonRelease:
+            warnx("got ButtonRelease (%u)", ev.xbutton.button);
             done = DONE;
             break;
         case KeyPress:
[scrot master]~> ./scrot -s - >/dev/null
scrot: got ButtonPress (1)
scrot: got ButtonRelease (3)
```

I was expecting to see `press 1, press 3, release 3` but the `press 3` event didn't come...

7xnl: While the warning does say "press", actually trying to match a press event might cause unwanted behavior.
The reason you're not seeing `press 3` is because the `scrotSelectionMotionDraw` function applies the following event mask:
```c
static void scrotSelectionMotionDraw(int x0, int y0, int x1, int y1)
{
    const struct Selection *const sel = &selection;
    const unsigned int EVENT_MASK = ButtonMotionMask | ButtonReleaseMask;
    ...
}
```
So once you start moving the mouse after starting the selection, clicking any other mouse buttons won't actually do anything (note that if you hold down the left mouse button, without moving the mouse, press events for other mouse buttons will get picked up).
The "fix" for this is adding the `ButtonPressMask` flag, but like I said, this might cause unwanted behavior. I'm not familiar enough with the codebase to know for sure.

Practically, it doesn't matter if what actually triggers the abort is a button press or release. If anything, making it trigger on a release event makes it more user friendly, because the user might want to cancel, but then while holding down the right mouse button decide to take the screenshot anyway.

I suppose a better warning message would be "Right mouse button was *clicked*, aborting"?

N-R-K:
I went with a slightly different patch (https://github.com/resurrecting-open-source-projects/scrot/commit/4976938970379b1db261687e921d07ebfebd8b54) since I didn't want to break the workflow for those who use right mouse to select.

PR #392, opened 2024-12-01
Add implementation of autoconf macro

zvr:
When trying to build on a new development machine, the build failed because the `PKG_CHECK_MODULES()` autoconf macro was not known.

This PR simply adds its definition in the standard way: as a file in a directory named `m4` and updates `./autogen.sh` to call `autoreconf` to use this directory. I've put the latest version of `pkg.m4` in the directory.

From a quick search on sources on my system, this is handled the same way in `gcc`, `gdb`, `binutils`, `grep`, `gnupg`, `freetype`, `mosh`, `GraphicsMagick` and many other projects.

Signed-off-by: Alexios Zavras (zvr) <zvr+github@zvr.gr>


zvr:
Ah, you are correct that it shouldn't be needed, if `pkg-config` is present. It's happening on my system because I have self-compiled `autotools` that only look at `/usr/local/share/aclocal` and not `/usr/share/aclocal`.

I assume then a check in `configure.ac` is needed, to ensure `pkg-config` is present.
Copying from the comments in the fie:

```
m4_ifndef([PKG_PREREQ],
     [m4_fatal([must install pkg-config 0.29 or later before running autoconf/autogen])])
```

N-R-K:
> I assume then a check in `configure.ac` is needed, to ensure `pkg-config` is present. Copying from the comments in the fie:
> 
> ```
> m4_ifndef([PKG_PREREQ],
>      [m4_fatal([must install pkg-config 0.29 or later before running autoconf/autogen])])
> ```

Sure, feel free to add that.

PR #396, opened 2025-01-02
ci: don't install unnecessary deps

N-R-K:
the ci script is referenced in CONTRIBUTING.md as an example, so keep dependencies to a minimal.

also use meta-package like "build-base" on alpine instead of installing binutils, make etc seperately.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -51,10 +51,9 @@ task:
       - CC: clang
       - CC: gcc
   install_script:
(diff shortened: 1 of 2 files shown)


N-R-K:
I noticed that there were couple warnings from github

image

Upgrading the `actions/checkout` version from `v2.4.0` to `v4.2.2` (latest) seems to make them go away.

If it's using semantic versioning scheme, then wouldn't it be better to do just `v4` instead of hardcoding a minor+patch version as well?

PR #406, opened 2025-03-07
replace xinerama with xrandr

N-R-K:
xrandr is supposedly the [better replacement](https://wiki.archlinux.org/title/Multihead#Xinerama) for xinerama. On my system, no packages (other than scrot) use xinerama while there's a bunch using xrandr, so it's likely a more common dependency which is probably already installed too.

Alternatively, instead of replacing it we could offer a build time config switch allowing the user to select between xrandr and xinerama.

### DIFF
--- a/deps.pc
+++ b/deps.pc
@@ -2,4 +2,4 @@ Name: scrot's mandatory dependencies
 Description: ditto
 Version: infinite
 Cflags: -D_XOPEN_SOURCE=700L
-Requires: x11 imlib2 >= 1.11.0 xcomposite >= 0.2.0 xext xfixes >= 5.0.1 xinerama >= 1.1.3
+Requires: x11 imlib2 >= 1.11.0 xcomposite >= 0.2.0 xext xfixes >= 5.0.1 xrandr >= 1.5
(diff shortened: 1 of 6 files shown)


N-R-K:
Anyone with multi monitor wants to try this out? The patch should be functional at it's current state. Though documentation and CI needs updating before merging - I'll do it later once the patch is confirmed to be good.

ping: @daltomi @zevweiss

daltomi:
It works correctly. I tested it with only one external monitor (VGA output).

Various configurations, from different resolutions, and specifying the primary monitor via `xrandr`.

```
scrot -M0 (ok, primary iterating via xrandr)
scrot -M1 (ok)
scrot -M2 (ok, error as expected)
```

deps.pc
```diff
@@ -2,4 +2,4 @@ Name: scrot's mandatory dependencies
 Description: ditto
 Version: infinite
 Cflags: -D_XOPEN_SOURCE=700L
-Requires: x11 imlib2 xcomposite xext xfixes xinerama
+Requires: x11 imlib2 xcomposite xext xfixes xrandr
```
daltomi, line 5: Remember to update the `README.md`

N-R-K:
@daltomi @zevweiss thanks for testing! I've merged it into `v2` branch where there's a couple other library related changes happening too.

PR #415, opened 2025-06-02
Removed focussed option and simplified zsh completion

DogeTheBeast:
Removed the alternate spelling of focused (ie focussed) to simplify the completion menu for zsh
Cleaned up the zsh completion menu by

- Removing the short options

### DIFF
--- a/etc/zsh-completion/_scrot
+++ b/etc/zsh-completion/_scrot
@@ -14,17 +14,17 @@ for ln in ${(f)list}; do
     local lopt="${tokens[2]}"
     local argtype=${tokens[3]%%:*}
     local argdesc=${tokens[3]#*:}
-    local desc="[${tokens[4]}]"
+    if [[ "${sopt}" = [[:alnum:]]* ]]; then 
+        local desc="[(-${sopt}) ${tokens[4]}]"
+    else 
+        local desc="[     ${tokens[4]}]"
+    fi
 
     case "$argtype" in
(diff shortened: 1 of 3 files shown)


etc/zsh-completion/_scrot
```diff
@@ -21,10 +21,12 @@ for ln in ${(f)list}; do
...
         N) ;; # None
     esac
-
     if [[ "${sopt}" = [[:alnum:]]* ]]; then
-        args+=( "-${sopt}${desc}" )
+        args+=("(-${sopt} --${lopt})-${sopt}${desc}")
+        args+=("(-${sopt} --${lopt})--${lopt}${desc}")
+    else 
+        args+=( "--${lopt}${desc}" )
     fi
```
N-R-K, line 29: This looks fine.

src/options.c
```diff
@@ -107,8 +107,6 @@ static const struct option lopts[] = {
     {"select",          optional_argument,  NULL,   's'},
     {"thumb",           required_argument,  NULL,   't'},
     {"focused",         no_argument,        NULL,   'u'},
-    /* macquarie dictionary has both spellings */
-    {"focussed",        no_argument,        NULL,   'u'},
```
N-R-K, line 111: This is unnecessary. You can just do the following inside the zsh completion to exclude it from the completion:

```zsh
    if [[ $lopt == focussed ]]; then
        continue
    fi
```

PR #417, opened 2025-08-22
Add --clipboard Flag

cwarden:
Add --clipboard flag to copy the full path to the screenshot to the clipboard.

### DIFF
--- a/man/scrot.1
+++ b/man/scrot.1
@@ -6,7 +6,7 @@
 .SH SYNOPSIS
 .nf
 .fam C
-\fBscrot\fP [\fB-bcfhimopuvz\fP] [\fB-a\fP \fIX,Y,W,H\fP] [\fB-C\fP \fINAME\fP] [\fB-D\fP \fIDISPLAY\fP] [\fB-d\fP \fISEC\fP] [\fB-e\fP \fICMD\fP]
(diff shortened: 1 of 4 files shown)


daltomi:
Hi, thanks for contributing.

Please do not modify the `man/scrot.1` file, only modify the `man/scrot.txt` file.

The idea is interesting, but I propose a solution (I think it would be simpler) using options already included in `scrot`, such as the execution option (`-e`) plus the use of the output format `SPECIAL STRINGS`.
Let me explain:
```bash
scrot -e 'echo $x | xclip -selection clipboard'
```
Where `$x` is the __new format__ with the full path.

There is a `$f` format, but the documentation is incorrect; it doesn't represent the __full path__, but rather the __full file name__.

Pseudocode:
- Implement a new format for `SPECIAL STRINGS`, proposal `$x`
- Use the format only if no output filename is specified (just like the `$f` format clarifies.)
- Concatenate `getcwd` with the full name of the output file.

Take it as just a proposal on my part, I don't have time right now to implement it.

N-R-K:
I think `--clipboard` is a bit confusing because I'd expect it to copy the *image* not the path. But IMO it's best not to get involved in the clipboard business and instead leave it to tools like `xclip` (or `xsel`).

So something like this already works:

```sh
scrot -e 'realpath $f | xclip -selection clipboard'
```

Adding a new format specifier that outputs the fully resolved path, like @daltomi suggested, is also okay and would remove the need to use `realpath`.

Issue #369, opened 2023-09-22
Selection via keyboard


N-R-K:
Since scrot is aimed at cli users, it's a fair assumption that a decent amount of scrot's users may prefer using the keyboard over mouse.

And so being able to select via keyboard could be a nice improvement.

hew02:
Just to confirm, when you say selection you are referring to window selection correct?

N-R-K:
> when you say selection you are referring to window selection correct?

No, area selection is possible with keyboard too. You can take a look at [selx](https://codeberg.org/NRK/selx) to see how it can be done if you're interested in working on this.

hew02:
Oh I see, nice! Yeah selx feels great. I'd be down to work on this, started using scrot and it's perfectly simple. So should this encompass both area selection and window select?

PR #427, opened 2026-08-19
Added selection via keyboard

hew02:
I wasn't sure if the ignore keyboard option, as it stands, should now be enabled by default, for now I have kept the original intent. Obviously borrows for `selx`. Let me know if there need to be further adjustments. #369.

### DIFF
--- a/man/scrot.1
+++ b/man/scrot.1
@@ -88,6 +88,14 @@ Display help and exit.
 Don't exit for keyboard input. ESC still exits.
 .TP
 .B
+\fB-K\fP, \fB--keyboard\fP
+Enables the use of selection via keyboard. Both
+hjkl and arrow keys are valid for navigation, escape 
+exits without taking a screenshot and space will 
+start the rectangle select mode. An additional 
+press of space to finalize a screenshot.
+.TP
+.B
 \fB-k\fP, \fB--stack\fP[=\fIOPT\fP]
 Capture stack/overlapped windows and join them. A
 running Composite Manager is needed for it to work
--- a/man/scrot.1
+++ b/man/scrot.1
@@ -93,7 +93,8 @@ Enables the use of selection via keyboard. Both
 hjkl and arrow keys are valid for navigation, escape 
 exits without taking a screenshot and space will 
 start the rectangle select mode. An additional 
-press of space to finalize a screenshot.
+press of space is then required to finalize a 
+screenshot.
 .TP
 .B
 \fB-k\fP, \fB--stack\fP[=\fIOPT\fP]
--- a/src/options.c
+++ b/src/options.c
@@ -79,7 +79,7 @@ enum { /* long opt only */
     OPT_FORMAT = UCHAR_MAX + 1,
     OPT_LIST_OPTS,
 };
-static const char stropts[] = "a:bC:cD:d:e:F:fhik::l:M:mopq:s::t:uvw:Z:z";
+static const char stropts[] = "a:bC:cD:d:e:F:fhiKk::l:M:mopq:s::t:uvw:Z:z";
 // NOTE: make sure lopts and opt_description indexes are kept in sync
 static const struct option lopts[] = {
     {"autoselect",      required_argument,  NULL,   'a'},
@@ -93,6 +93,7 @@ static const struct option lopts[] = {
     {"freeze",          no_argument,        NULL,   'f'},
     {"help",            no_argument,        NULL,   'h'},
     {"ignorekeyboard",  no_argument,        NULL,   'i'},
+    {"usekeyboard",     no_argument,        NULL,   'K'},
     {"stack",           optional_argument,  NULL,   'k'},
     {"line",            required_argument,  NULL,   'l'},
     {"monitor",         required_argument,  NULL,   'M'},
@@ -128,6 +129,7 @@ static const struct option_desc {
     /* f */  { "freeze the screen when -s is used", "" },
     /* h */  { "display help and exit", "" },
     /* i */  { "ignore keyboard", "" },
+    /* K */  { "enable keyboard control", "" },
     /* k */  { "capture overlapped window and join them", "v|h" },
     /* l */  { "specify the style of the selection line", "STYLE" },
     /* M */  { "capture monitor", "NUM" },
@@ -411,6 +413,9 @@ void optionsParse(int argc, char *argv[])
             break;
         case 'i':
             opt.ignoreKeyboard = true;
+            break;
+        case 'K':
+            opt.useKeyboard = true;
             break;
         case 'k':
             opt.mode = MODE_STACK;
--- a/src/scrot.c
+++ b/src/scrot.c
@@ -107,9 +107,14 @@ int main(int argc, char *argv[])
     /* Get the time ASAP to reduce the timing error in case --delay is used. */
     opt.delayStart = clockNow();
 
-    atexit(uninitXAndImlib);
-
     optionsParse(argc, argv);
+
+    if (opt.ignoreKeyboard && opt.useKeyboard) {
+      fprintf(stderr, "scrot: can not read input from keyboard if also set to be ignored\n");
+      return 1;
+    }
+
+    atexit(uninitXAndImlib);
 
     initXAndImlib(opt.display, 0);
 
--- a/src/scrot_selection.c
+++ b/src/scrot_selection.c
@@ -206,6 +206,7 @@ static bool scrotSelectionGetUserSel(struct SelectionRect *selectionRect)
     enum { WAIT, DONE, ABORT } done = WAIT;
     int rx = 0, ry = 0, rw = 0, rh = 0;
     bool isButtonPressed = false;
+    bool isStartSelectionKeyPressed = false;
     unsigned int buttonId = -1;
     Window target = None;
     Status ret;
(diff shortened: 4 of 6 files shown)


N-R-K:
How is this supposed to work? I'm trying `scrot --select` but pressing arrow key just terminates the shot:

```
scrot: Key was pressed, aborting shot
scrot: no image grabbed
```

hew02:
> How is this supposed to work? I'm trying `scrot --select` but pressing arrow key just terminates the shot:
> 
> ```
> scrot: Key was pressed, aborting shot
> scrot: no image grabbed
> ```

That's what I was referring to in my above comment, maybe not clear enough. As it is, scrot requires an additional flag to not exit with keyboard input: -i/--ignorekeyboard. Didn't want to touch things I wasn't directly working on but I wonder if it is in fact necessary.

N-R-K:
> scrot requires an additional flag to not exit with keyboard input: -i/--ignorekeyboard.

I see, that's certainly confusing. I'd expect ignore keyboard to... well ignore keyboard. I suggest adding a separate --keyboard/-K flag to enable keyboard control.

* No flags: default behavior.
* `-i`: follow existing behavior with `-i` and ignore keyboard.
* `-K`: allow keyboard control (new behavior).
* `-i` + `-K`: mutually exclusive, write an error message and exit.

hew02:
I was wondering if diagonal movement would be nice as well?

man/scrot.1
daltomi, line 1: Please do not modify this file, only the `man/scrot.txt` file.
This file is modified by the maintainer at the time of the new release.

daltomi:
This PR would break the current behavior of being able to correct the **start of the selection** via keyboard and mouse, which is the most important part, in my opinion.

In other words, in the **master branch** you can correct the start of the selection:

https://github.com/user-attachments/assets/3bad0328-6db1-478e-81e4-f7831aca5942

man/scrot.1
hew02: Sorry, missed that!

hew02:
> This PR would break the current behavior of being able to correct the **start of the selection** via keyboard and mouse, which is the most important part, in my opinion.
> 
> In other words, in the **master branch** you can correct the start of the selection:
> scrot_select_master.mp4

Fixing that on my end, big oversight on my end, thanks. Should that functionality also be present in keyboard controls? Say holding alt will correct the start?

daltomi:
Yes, I think it should work, that is, we'll have to test it thoroughly.

hew02:
Thank you for the feedback thus far

src/scrot_selection.c
```diff
@@ -253,44 +254,94 @@ static bool scrotSelectionGetUserSel(struct SelectionRect *selectionRect)
...
+                    if (!isStartSelectionKeyPressed) {
+                        isStartSelectionKeyPressed = true;
+                        rx = p.x; ry = p.y;
+                    }
+                    else {
+                        done = DONE;
+                    }
+                    break;
+                }
             default:
```
daltomi, line 317: Hi,
I'm getting this compiler warning. Sometimes it's correct, sometimes it's not; it depends on what you're trying to do here.
```bash
scrot_selection.c: In function ‘scrotSelectionGetUserSel’:
scrot_selection.c:303:20: warning: this statement may fall through [-Wimplicit-fallthrough=]
  303 |                 if (opt.useKeyboard) {
      |                    ^
scrot_selection.c:317:13: note: here
  317 |             default:
      |             ^~~~~~~
```

If you're going to keep it, you should indicate this with a comment (// fallthrough) in the source code (it's standard practice). This disables the warning from some compilers, such as GCC at least.

```c
  }
 // fallthrough
default:
...
```

man/scrot.txt
```diff
@@ -41,6 +41,12 @@ OPTIONS
   -f, --freeze              Freeze the screen when -s is used.
   -h, --help                Display help and exit.
   -i, --ignorekeyboard      Don't exit for keyboard input. ESC still exits.
+  -K, --keyboard            Enables the use of selection via keyboard. Both
```
daltomi, line 44: I've noticed some trailing spaces in this and other files; it would be good to correct them. You can see them with:
```bash
git diff --check master
```

Image

daltomi:
I encountered a strange situation; if I run the command `-s -K` with the corresponding parameters, the crosshair cursor appears, waiting for the user to move it with the assigned keys—all good so far. But if, instead, I press the **Alt key first** and then the cursor movement keys, I can no longer capture the image (by pressing the spacebar).

In short:
1) `scrot -s -K`
2) Press Alt + up/down/etc. and release. (note how I didn't press the space bar to start the selection; I skipped that.)

3) Press the spacebar (first time)
4) Capture fails.

The solution to this behavior would be to **ignore the Alt key** until the **first selection is made**.

I'm not sure if I've explained myself clearly :-)

daltomi (changes requested):

src/scrot_selection.c
daltomi: News:
I hadn't noticed, but GitHub Actions is failing to compile due to this warning.

hew02:
> I encountered a strange situation; if I run the command `-s -K` with the corresponding parameters, the crosshair cursor appears, waiting for the user to move it with the assigned keys—all good so far. But if, instead, I press the **Alt key first** and then the cursor movement keys, I can no longer capture the image (by pressing the spacebar).
> 
> In short:
> 
>     1. `scrot -s -K`
> 
>     2. Press Alt + up/down/etc. and release. (note how I didn't press the space bar to start the selection; I skipped that.)
> 
>     3. Press the spacebar (first time)
> 
>     4. Capture fails.
> 
> 
> The solution to this behavior would be to **ignore the Alt key** until the **first selection is made**.
> 
> I'm not sure if I've explained myself clearly :-)

Makes perfect sense and I'm going through and trying different combinations of possible inputs. Is it worth keeping the distinction between a button press and a keyboard key to start the selection? As I see it what matters is that a selection has started and some behavior should be gated until that occurs.

src/scrot_selection.c
hew02: Should have left a comment! Strange that Clang doesn't report that...but yes kept the fallthrough for when keyboard input isn't enabled.

src/scrot.c
```diff
@@ -107,10 +107,15 @@ int main(int argc, char *argv[])
     /* Get the time ASAP to reduce the timing error in case --delay is used. */
     opt.delayStart = clockNow();
 
-    atexit(uninitXAndImlib);
```
N-R-K, line 110: Why was this removed? Seems unintentional.

hew02: My intent was to first check for the discrepancy between ignoring and using the keyboard then register the atexit, however, looking again a the uninit function seems not super necessary. I'll revert it to as it was.

src/scrot_selection.c
```diff
@@ -253,44 +254,97 @@ static bool scrotSelectionGetUserSel(struct SelectionRect *selectionRect)
         {
             KeySym *keysym = NULL;
             int keycode; /*dummy*/
+			      struct Point p = { ev.xkey.x, ev.xkey.y };
```
daltomi: I stumbled upon these tab characters. They need to be removed to maintain the project's code style.

daltomi (approved):

src/options.c
```diff
@@ -93,6 +93,7 @@ static const struct option lopts[] = {
     {"freeze",          no_argument,        NULL,   'f'},
     {"help",            no_argument,        NULL,   'h'},
     {"ignorekeyboard",  no_argument,        NULL,   'i'},
+    {"usekeyboard",     no_argument,        NULL,   'K'},
```
N-R-K, line 96: Manpage says `--keyboard` but here it's `--usekeyboard`. I prefer the former.

N-R-K (changes requested):

daltomi:
> Is it worth keeping the distinction between a button press and a keyboard key to start the selection? As I see it what matters is that a selection has started and some behavior should be gated until that occurs.

I’m not sure if I understood you correctly, but the idea is that if the user employs the `--select` option alone, the selection should be initiated using the mouse button.
If the `-K` option is also used, the selection should start by pressing the space bar.
In other words, if the `-K` option is used, the mouse should not trigger any selection; the cursor moves, but that is all.

Something like that.

```patch
diff --git a/src/scrot_selection.c b/src/scrot_selection.c
index 71fe6a0..d376a9e 100644
--- a/src/scrot_selection.c
+++ b/src/scrot_selection.c
@@ -226,6 +226,9 @@ static bool scrotSelectionGetUserSel(struct SelectionRect *selectionRect)
     }
 
     while (done == WAIT) {
+        if (opt.useKeyboard)
+            isButtonPressed = false;
+
         XNextEvent(disp, &ev);
         switch (ev.type) {
         case MotionNotify:
@@ -338,7 +341,7 @@ static bool scrotSelectionGetUserSel(struct SelectionRect *selectionRect)
                             scrotSelectionMotionDraw(rx, ry, p.x, p.y);
                     }
                 }
-                else {
+                else if (isButtonPressed) {
                     rx = r.x; ry = r.y;
                     scrotSelectionMotionDraw(rx, ry, ev.xkey.x, ev.xkey.y);
                 }
```
:-) That is just my opinion, of course; if you have other ideas, we can try them out.

hew02:
> > Is it worth keeping the distinction between a button press and a keyboard key to start the selection? As I see it what matters is that a selection has started and some behavior should be gated until that occurs.
> 
> I’m not sure if I understood you correctly, but the idea is that if the user employs the `--select` option alone, the selection should be initiated using the mouse button. If the `-K` option is also used, the selection should start by pressing the space bar. In other words, if the `-K` option is used, the mouse should not trigger any selection; the cursor moves, but that is all.
> 
> Something like that.
> 
> ```diff
> diff --git a/src/scrot_selection.c b/src/scrot_selection.c
> index 71fe6a0..d376a9e 100644
> --- a/src/scrot_selection.c
> +++ b/src/scrot_selection.c
> @@ -226,6 +226,9 @@ static bool scrotSelectionGetUserSel(struct SelectionRect *selectionRect)
>      }
>  
>      while (done == WAIT) {
> +        if (opt.useKeyboard)
> +            isButtonPressed = false;
> +
>          XNextEvent(disp, &ev);
>          switch (ev.type) {
>          case MotionNotify:
> @@ -338,7 +341,7 @@ static bool scrotSelectionGetUserSel(struct SelectionRect *selectionRect)
>                              scrotSelectionMotionDraw(rx, ry, p.x, p.y);
>                      }
>                  }
> -                else {
> +                else if (isButtonPressed) {
>                      rx = r.x; ry = r.y;
>                      scrotSelectionMotionDraw(rx, ry, ev.xkey.x, ev.xkey.y);
>                  }
> ```
> 
> :-) That is just my opinion, of course; if you have other ideas, we can try them out.

Yes that makes sense to me, my question, if I remember correctly, is stemming from the possibility to start with mouse select and begin an area select then pressing space would start a keyboard select and create a new origin point. The opposite is true as well. My question then being should there only be on one chance to create a new origin point, either keyboard or mouse?

hew02:
> > > Is it worth keeping the distinction between a button press and a keyboard key to start the selection? As I see it what matters is that a selection has started and some behavior should be gated until that occurs.
> > 
> > 
> > I’m not sure if I understood you correctly, but the idea is that if the user employs the `--select` option alone, the selection should be initiated using the mouse button. If the `-K` option is also used, the selection should start by pressing the space bar. In other words, if the `-K` option is used, the mouse should not trigger any selection; the cursor moves, but that is all.
> > Something like that.
> > ```diff
> > diff --git a/src/scrot_selection.c b/src/scrot_selection.c
> > index 71fe6a0..d376a9e 100644
> > --- a/src/scrot_selection.c
> > +++ b/src/scrot_selection.c
> > @@ -226,6 +226,9 @@ static bool scrotSelectionGetUserSel(struct SelectionRect *selectionRect)
> >      }
> >  
> >      while (done == WAIT) {
> > +        if (opt.useKeyboard)
> > +            isButtonPressed = false;
> > +
> >          XNextEvent(disp, &ev);
> >          switch (ev.type) {
> >          case MotionNotify:
> > @@ -338,7 +341,7 @@ static bool scrotSelectionGetUserSel(struct SelectionRect *selectionRect)
> >                              scrotSelectionMotionDraw(rx, ry, p.x, p.y);
> >                      }
> >                  }
> > -                else {
> > +                else if (isButtonPressed) {
> >                      rx = r.x; ry = r.y;
> >                      scrotSelectionMotionDraw(rx, ry, ev.xkey.x, ev.xkey.y);
> >                  }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > :-) That is just my opinion, of course; if you have other ideas, we can try them out.
> 
> Yes that makes sense to me, my question, if I remember correctly, is stemming from the possibility to start with mouse select and begin an area select then pressing space would start a keyboard select and create a new origin point. The opposite is true as well. My question then being should there only be on one chance to create a new origin point, either keyboard or mouse?

Sorry, just reading your response again and realizing you answered my question haha!

